PR #20, opened 2021-04-16
Update bindgen

willbrowningme:
This pull request updates the bindgen version in build-depenecies and allows it to have a range of versions.

It also fixes a failing test that was caused by the OUTER_LINE_DOC comments.

I had been using the https://github.com/killertux/solder repo but it appears abandonded and lacks php8 support. So it's great to see that you are working on this!

### DIFF
--- a/src/php/types/zval.rs
+++ b/src/php/types/zval.rs
@@ -72,8 +72,8 @@ impl<'a> Zval {
             // We can safely cast our *const c_char into a *const u8 as both
             // only occupy one byte.
             unsafe {
-                let len = (*self.value.str).len;
-                let ptr = (*self.value.str).val.as_ptr() as *const u8;
+                let len = (*self.value.str_).len;
+                let ptr = (*self.value.str_).val.as_ptr() as *const u8;
                 let _str = std::str::from_utf8(slice::from_raw_parts(ptr, len as usize)).unwrap();
 
                 Some(_str.to_string())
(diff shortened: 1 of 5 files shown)


Lost-Packets:
@davidcole1340 Thoughts?

willbrowningme:
Hmm, not sure why this is failing. 

Tests are passing and it's building successfully on my Ubuntu 20.04 machine.

davidcole1340:
Thanks for the PR! Yeah, the tests run fine for me as well. I'm just gonna push a commit that will fix the doc comment (I forgot to rewrite when I refactored the table stuff), and I will have a look at the tests.

src/php/types/zval.rs
```diff
@@ -166,7 +166,7 @@ impl<'a> Zval {
 
             for param in params {
```
Lost-Packets, line 167: I like this part

davidcole1340 (approved):
I'll merge this for now and try to figure out the CI later, cheers!

davidcole1340:
Not sure why but we had to install LLVM and Clang, don't know why it was working before but all sorted now.

willbrowningme:
Awesome, thanks!

It's because bindgen has [clang-sys](https://github.com/KyleMayes/clang-sys) as a dependency - https://github.com/rust-lang/rust-bindgen/blob/master/Cargo.toml#L52

I shouldn't have just assumed those were already installed!

davidcole1340:
Yeah it seems that bindgen 0.53.1 used clang-sys 0.28.0, while the upgraded bindgen uses clang-sys 1.0.

PR #27, opened 2021-04-22
Relicense under MIT or Apache 2.0

davidcole1340:
To be consistent with other Rust crates, relicensed adding the Apache 2.0 license.

@willbrowningme since you have contributed to this project, can you please confirm that this is alright with you?


willbrowningme:
Yes happy with that. 

I'm using this crate here - [https://gitlab.com/willbrowning/anonaddy-sequoia](https://gitlab.com/willbrowning/anonaddy-sequoia).

Would you recommend changing that to Apache 2.0 too?

davidcole1340:
Awesome, thank you.

Up to you really - it seems that the Rust community mainly dual-licenses under MIT and Apache 2.0: https://rust-lang.github.io/api-guidelines/necessities.html

I don't really think it affects much but I'm not a lawyer haha

PR #36, opened 2021-05-06
Remove default-features=false from bindgen

willbrowningme:
Remove default-features=false from bindgen build dependency to fix failing builds with crates that have multiple different dependencies that both depend on bindgen.

See failed pipeline here for the error in question - https://gitlab.com/willbrowning/anonaddy-sequoia/-/jobs/1238041327

I believe this may be related to this issue - https://github.com/rust-lang/rust-bindgen/issues/2030


davidcole1340:
Cheers!

PR #63, opened 2021-09-07
Naively change the case of method identifiers to comply with PSR-1

vodik:
I'd be surprised if the patch will be taken in this form, but I'm willing to put the effort in to getting this in should there be interest.

Rust loves using snake_case for functions, but to comply with PSR-1, methods should be in camelCase. This patch tries to change the case of methods generated through the `php_impl` macro.

As for precedence for changing the case of identifiers via macros, both juniper and async-graphql do this.

### DIFF
--- a/ext-php-rs-derive/src/method.rs
+++ b/ext-php-rs-derive/src/method.rs
@@ -1,4 +1,5 @@
 use anyhow::{anyhow, bail, Result};
+use ident_case::RenameRule;
 use quote::ToTokens;
 use std::collections::HashMap;
 
@@ -32,6 +33,29 @@ pub struct Method {
     pub output: Option<(String, bool)>,
     pub _static: bool,
     pub visibility: Visibility,
+}
+
+fn camel_case_identifier(field: &str) -> String {
+    match field {
+        "__construct" => "__construct".to_string(),
+        "__destruct" => "__destruct".to_string(),
+        "__call" => "__call".to_string(),
+        "__call_static" => "__callStatic".to_string(),
+        "__get" => "__get".to_string(),
+        "__set" => "__set".to_string(),
+        "__isset" => "__isset".to_string(),
+        "__unset" => "__unset".to_string(),
+        "__sleep" => "__sleep".to_string(),
+        "__wakeup" => "__wakeup".to_string(),
+        "__serialize" => "__serialize".to_string(),
+        "__unserialize" => "__unserialize".to_string(),
+        "__to_string" => "__toString".to_string(),
+        "__invoke" => "__invoke".to_string(),
+        "__set_state" => "__set_state".to_string(),
+        "__clone" => "__clone".to_string(),
+        "__debug_info" => "__debugInfo".to_string(),
+        field => RenameRule::CamelCase.apply_to_field(field),
+    }
 }
 
 pub fn parser(input: &mut ImplItemMethod) -> Result<(TokenStream, Method)> {
@@ -93,7 +117,7 @@ pub fn parser(input: &mut ImplItemMethod) -> Result<(TokenStream, Method)> {
     };
 
     let method = Method {
-        name: ident.to_string(),
+        name: camel_case_identifier(&ident.to_string()),
         ident: internal_ident.to_string(),
         args,
         optional,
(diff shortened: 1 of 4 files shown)


ext-php-rs-derive/src/method.rs
```diff
@@ -92,8 +93,14 @@ pub fn parser(input: &mut ImplItemMethod) -> Result<(TokenStream, Method)> {
         }
     };
 
+    let name = if ident == "__construct" {
```
Lctrs, line 96: We should apply this to all magic methods.

vodik: Oh that's a good point. How would we handle `__toString()`?

Should we just strip the double underscore prefix, normalize the token, then reattach it? In which case `__to_string` would become `__toString`.

Lctrs: Have you tried to to just pass `__to_string` to `ident_case` to see how it goes ?

vodik: Right, sorry. Should have explained that the reason for '__construct' special case is it got converted to 'construct'. I t doesn't seem like `indent_case` respects leading underscores.

vodik:
Okay, did a bit more research and looks like this is something darling should handle for us.

vodik:
I didn't realize that PHP itself isn't consistent with the casing of magic methods (`__set_state` is an exception) so I figured it's probably safest to just hardcode in a list.

davidcole1340:
I think this could be useful, but there should probably be a way to opt-in or opt-out using an attribute. I can help you with that if you want (it's basically the same as the rest of the impl attributes, see [here]).

Let me know what you guys think.

[here]: https://github.com/mcquaig-institute/ext-php-rs/blob/naive-psr1/ext-php-rs-derive/src/method.rs#L66-L72

ext-php-rs-derive/src/method.rs
```diff
@@ -34,6 +35,29 @@ pub struct Method {
     pub visibility: Visibility,
 }
```
davidcole1340, line 37: Maybe a comment just describing the function?

ext-php-rs-derive/src/method.rs
```diff
@@ -93,7 +117,7 @@ pub fn parser(input: &mut ImplItemMethod) -> Result<(TokenStream, Method)> {
     };
 
     let method = Method {
-        name: ident.to_string(),
+        name: camel_case_identifier(&ident.to_string()),
```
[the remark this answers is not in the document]
davidcole1340: Nevermind, `ident` is `Ident`.

vodik:
@davidcole1340 I'll give it a shot. Its a good idea.

vodik:
Added an attribute that allows for some limited control over casing. Only implemented camelCase and snake_case:
```rust 
#[php_impl]
    #[rename("snake_case")]
    pub fn get_first_name(&self) -> String {
        // ...
    }

    #[rename("camelCase")]
    pub fn get_last_name(&self) -> String {
        // ...
    }
}
```
Looks like `ident_name` can't properly convert camel case identifiers into snake case though. I don't know how much of a problem this is in practice. I picked this library only because its already pulled in by another dep, but there are other implementations.

If this is acceptable, let me know and I'll put a blurb in the docs as well.

Lctrs:
I honestly don't like the naming. A new user will think that their method will be renamed to literally `snake_case` or `camelCase` at first glance.

Also, I would have preferred a global attribute for that. I doubt that users will want to switch case to only a few methods.

vodik:
Honestly, I think I agree. Not sure how a global attribute would look as I don't think I've come across a library where I've used them.

But as I was putting this together I was thinking just having rename be a bit more straightforward could be better: `#[rename("my_identifier_of_choice")]` instead of using it to control how to mangle the identifier.

davidcole1340:
Could also do `#[rename(snake)]` and `#[rename(camel)]`? What do you think?

vodik:
I kept it with "snake_case" and "camcelCase" to keep it consistent with other rust packages (e.g. [serde](https://serde.rs/attr-rename.html), [async-graphql](https://docs.rs/async-graphql/2.9.15/async_graphql/attr.Object.html)).

That said, as much as I've just appealed to how others do this, they don't do it with a `rename` attribute per say. I think what could work:

- Put a `rename_methods` parameter on `php_impl`: `#[php_impl(rename_methods="snake_case")]`
- Change `rename` to directly take an identifier and allow directly overridding the exported name of a function.

davidcole1340:
Thanks!

PR #67, opened 2021-09-14
Add wrapper for zend_exec_stringl with exception handling

vodik:
Add a wrapper around eval. This function exposes zend's behaviour (expects an expression which is automatically prepended with "return"), which is different PHP's builtin's behaviour.

I'm not sure if it would make sense to make a PHP equivalent `eval` as well (should be really easy to make), so I don't know if we want to be conservative with the exported name.

The eval portion works nicely, but I have no idea how to go from `ZendObject` to `PHPException`. Currently working around this by shoving the `ZendObject` into `Error`, but this feels icky.


src/php/globals.rs
```diff
@@ -16,6 +16,14 @@ impl ExecutorGlobals {
             .expect("Static executor globals were invalid")
     }
 
+    fn get_mut() -> &'static mut Self {
+        // SAFETY: PHP executor globals are statically declared therefore should never
+        // return an invalid pointer.
+        // TODO: Should this be syncronized?
+        unsafe { ext_php_rs_executor_globals().as_mut() }
+            .expect("Static executor globals were invalid")
```
vodik, line 24: Kept this private because I don't know in general if this is thread safe. As I understand it, the `take_exception` function should be safe.

src/php/globals.rs
```diff
@@ -24,4 +32,17 @@ impl ExecutorGlobals {
...
     }
+
+    pub fn take_exception() -> Option<Box<ZendObject>> {
+        let globals = Self::get_mut();
+
+        let mut exception_ptr = std::ptr::null_mut();
+        std::mem::swap(&mut exception_ptr, &mut globals.exception);
+
+        if !exception_ptr.is_null() {
+            Some(unsafe { Box::from_raw(exception_ptr) })
```
vodik, line 43: As I understand the php source code, we're responsible for freeing it if we take it.

src/errors.rs
```diff
@@ -1,19 +1,20 @@
...
+    types::object::ZendObject,
 };
 
 /// The main result type which is passed by the library.
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// The main error type which is passed by the library inside the custom
 /// [`Result`] type.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone)]
```
vodik, line 17: By the way, what's the benefit of having ordering on an error enum?

vodik:
Example in use:
```rust
#[php_function]
pub fn example_eval() -> Result<Vec<String>> {
    ext_php_rs::php::eval::eval("['a', 'b', 'c']").unwrap().extract()
}
```

vodik:
Also I suspect this might be a bug, but it seems like wrapped functions are unable to directly return zvals? `pub fn example_eval() -> Result<Zval>` won't compile

src/errors.rs
kakserpom: Comparison.

ptondereau:
Hey @vodik, thanks for this contribution! This functionality has since been implemented in the codebase — we now have `src/php_eval.rs` with a more robust eval path using `zend_compile_string` + `zend_execute`, and `ExecutorGlobals::take_exception()` is also covered. Closing as superseded. Cheers!

PR #74, opened 2021-09-20
Add Display impl for ZendObject

vodik:
Small PR that brings part of my work getting eval/compile support built.

Adds some basic support for capturing PHP exceptions, function calling, and convenience methods around converting from Zval to Rust types.

Which that groundwork in place, add a PHP equivalent implementation for Display.

### DIFF
--- a/src/php/globals.rs
+++ b/src/php/globals.rs
@@ -2,7 +2,7 @@
 
 use crate::bindings::{_zend_executor_globals, ext_php_rs_executor_globals};
 
-use super::types::array::HashTable;
+use super::types::{array::HashTable, object::ZendObject};
 
 /// Stores global variables used in the PHP executor.
 pub type ExecutorGlobals = _zend_executor_globals;
@@ -16,8 +16,32 @@ impl ExecutorGlobals {
             .expect("Static executor globals were invalid")
     }
 
+    fn get_mut() -> &'static mut Self {
+        // SAFETY: PHP executor globals are statically declared therefore should never
+        // return an invalid pointer.
+        // TODO: Should this be syncronized?
+        unsafe { ext_php_rs_executor_globals().as_mut() }
+            .expect("Static executor globals were invalid")
+    }
+
     /// Attempts to retrieve the global class hash table.
     pub fn class_table(&self) -> Option<&HashTable> {
         unsafe { self.class_table.as_ref() }
+    }
+
+    /// Attempts to extract the last PHP exception captured by the interpreter.
+    ///
+    /// Note that the caller is responsible for freeing the memory here or it'll leak.
+    pub fn take_exception() -> Option<*mut ZendObject> {
+        let globals = Self::get_mut();
+
+        let mut exception_ptr = std::ptr::null_mut();
+        std::mem::swap(&mut exception_ptr, &mut globals.exception);
+
+        if !exception_ptr.is_null() {
+            Some(exception_ptr)
+        } else {
+            None
+        }
     }
 }
--- a/src/php/types/object.rs
+++ b/src/php/types/object.rs
@@ -17,10 +17,11 @@ use std::{
 use crate::{
     bindings::{
         ext_php_rs_zend_object_alloc, ext_php_rs_zend_object_release, object_properties_init,
-        std_object_handlers, zend_is_true, zend_object, zend_object_handlers, zend_object_std_dtor,
-        zend_object_std_init, zend_objects_clone_members, zend_std_get_properties,
-        zend_std_has_property, zend_std_read_property, zend_std_write_property, zend_string,
-        HashTable, ZEND_ISEMPTY, ZEND_PROPERTY_EXISTS, ZEND_PROPERTY_ISSET,
+        std_object_handlers, zend_call_known_function, zend_is_true, zend_object,
+        zend_object_handlers, zend_object_std_dtor, zend_object_std_init,
+        zend_objects_clone_members, zend_std_get_properties, zend_std_has_property,
+        zend_std_read_property, zend_std_write_property, zend_string, HashTable, ZEND_ISEMPTY,
+        ZEND_PROPERTY_EXISTS, ZEND_PROPERTY_ISSET,
     },
     errors::{Error, Result},
     php::{
@@ -28,6 +29,7 @@ use crate::{
         enums::DataType,
         exceptions::PhpResult,
         flags::ZvalTypeFlags,
+        globals::ExecutorGlobals,
         types::{array::OwnedHashTable, string::ZendString},
     },
 };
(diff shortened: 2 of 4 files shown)


src/php/types/object.rs
```diff
@@ -161,6 +164,38 @@ impl ZendObject {
...
+                self.ce,
+                &mut ret,
+                0,
+                ptr::null_mut(),
+                ptr::null_mut(),
+            );
+        }
+
+        if let Some(err) = ExecutorGlobals::take_exception() {
+            let class_name = unsafe { ZendString::from_ptr((*self.ce).name, false) };
```
vodik, line 183: Passing `false` here feels weird. Would it maybe make sense to separate this into two type? A `ZendString` and `ZendStr` for example.

```rust
let class_name = unsafe { ZendStr::from_ptr((*self.ce).name) };

Comment (davidcole1340) [reply src/php/types/object.rs:183]: Yeah I agree, I feel like I've got the FFI wrappers wrong (strings and arrays). What do you think about doing the same for strings? See [here](https://github.com/davidcole1340/ext-php-rs/blob/3646da741caf04849b3552915af34f68ee16eac7/src/php/types/array.rs) for progress with the array.

Implement traits on references to the underlying `zend_string`, and then have an `OwnedZendString` when you need to create one?

File: src/php/types/object.rs [hunk 2]
```diff
@@ -161,6 +164,38 @@ impl ZendObject {
...
+            panic!(
+                "Uncaught exception during call to {}::__toString(): {:?}",
+                class_name.unwrap().as_str().unwrap(),
```
davidcole1340, line 186: These `unwrap()` errors are OK if you are panicing, will just need to disable the lint for the function.

src/php/types/object.rs
```diff
@@ -161,6 +164,38 @@ impl ZendObject {
     pub(crate) fn refcount_inc(&mut self) {
         self.gc.refcount += 1;
     }
+
+    pub(crate) fn to_string(&self) -> String {
```
davidcole1340, line 168: Should probably change function name eventually, although I'm not sure what to...

src/php/globals.rs
```diff
@@ -16,6 +16,14 @@ impl ExecutorGlobals {
             .expect("Static executor globals were invalid")
     }
 
+    fn get_mut() -> &'static mut Self {
+        // SAFETY: PHP executor globals are statically declared therefore should never
+        // return an invalid pointer.
+        // TODO: Should this be syncronized?
```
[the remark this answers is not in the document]
davidcole1340: Just a side note, ZTS is _technically_ supported, but I doubt that the library is thread-safe at the moment. I think proper ZTS support should be a goal, and I would eventually like to try some things with async.

src/php/types/object.rs
davidcole1340: I've changed the implementation of Zend strings, have a look [here](https://github.com/davidcole1340/ext-php-rs/commit/7ba9ee9bfb5d2888f6e83129aa4361cbb7cb43c5).

src/php/globals.rs
```diff
@@ -24,4 +32,17 @@ impl ExecutorGlobals {
...
     }
+
+    pub fn take_exception() -> Option<Box<ZendObject>> {
+        let globals = Self::get_mut();
+
+        let mut exception_ptr = std::ptr::null_mut();
+        std::mem::swap(&mut exception_ptr, &mut globals.exception);
+
+        if !exception_ptr.is_null() {
+            Some(unsafe { Box::from_raw(exception_ptr) })
```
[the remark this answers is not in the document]
vodik: Oh of course. I knew that looked wrong. Yeah, its just an owned Zval. That simplifies this a lot.

src/php/globals.rs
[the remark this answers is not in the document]
vodik: I'm fine with the suggestion, but the concern I have is PHP's internals don't seem to do any synchronization either. So it doesn't matter if we lock our access to it, all bets could be off depending on what PHP behind the scenes is doing.

src/php/types/object.rs
vodik: If you want to keep the lint, I can change them to expect. I don't intend for these to stay panics forever. PHP recovers from both situations. I'll put in comments when I revisit this.

But in order to remove the panics, first I need a sensible way to turn a Zval into an Error.

src/php/types/object.rs
vodik: `strval`?

src/php/types/object.rs
vodik: Looks promising, but haven't had a chance to deep dive into it. Will hopefully be able to tonight, no promises.

src/php/globals.rs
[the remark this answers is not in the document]
vodik: Also that's part of the reason I swap pointer. My understanding is that should be atomic.

src/php/types/object.rs
davidcole1340: Could do `as_string`, since `to_` usually implies consuming `self`.

vodik: Aren't `as_` functions supposed to be cheap though? This invokes arbitrary PHP.

vodik: Also, for what its worth, `to_` should be fine idiomatically for this: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

Consider the `str::to_lowercase` as an example.

davidcole1340: Yeah true, `strval` sounds good then?

vodik:
Okay, so I think there's a lot of good stuff in here that should get merged in to get iterated on:
- `Zval::extract` as a QoL improvment
- `ExecutorGlobals::take_exception`
- `FromZendObject` for implementing casts from objects.

That said, with all this built, I also think the `Display` implementation is inappropriate given that it can panic and I think should be dropped from this PR outright.

Thoughts?

src/php/globals.rs
```diff
@@ -16,8 +16,32 @@ impl ExecutorGlobals {
...
+
     /// Attempts to retrieve the global class hash table.
     pub fn class_table(&self) -> Option<&HashTable> {
         unsafe { self.class_table.as_ref() }
     }
+
+    /// Attempts to extract the last PHP exception captured by the interpreter.
+    ///
+    /// Note that the caller is responsible for freeing the memory here or it'll leak.
+    pub fn take_exception() -> Option<*mut ZendObject> {
```
davidcole1340, line 35: Return `*mut ZendObject` or load the pointer into a `Zval`, returning that? Or potentially `&'static ZendObject` (if it is non-null and well aligned)?

davidcole1340: Returning a `Zval` just gives it a nice wrapper which will drop the object when `Zval` goes out of scope.

src/php/types/object.rs
```diff
@@ -157,6 +159,65 @@ impl ZendObject {
...
+                0,
+                std::ptr::null_mut(),
+                std::ptr::null_mut(),
+            );
+        }
+
+        if let Some(err) = ExecutorGlobals::take_exception() {
+            // TODO: become an error
+            let class_name = obj.get_class_name();
+            panic!(
```
davidcole1340, line 205: Maybe change to errors before merging?

src/php/types/object.rs
```diff
@@ -177,6 +238,16 @@ impl Debug for ZendObject {
     }
 }
 
+impl fmt::Display for ZendObject {
```
davidcole1340, line 241: When does this panic?

src/php/types/zval.rs
```diff
@@ -427,6 +427,16 @@ impl Zval {
         unsafe { zval_ptr_dtor(self) };
         self.u1.type_info = ty.bits();
     }
+
+    /// Extracts some type from a `Zval`.
+    ///
+    /// This is a wrapper function around `TryFrom`.
+    pub fn extract<T>(self) -> Result<T>
+    where
+        T: TryFrom<Self, Error = Error>,
```
davidcole1340, line 436: `TryFrom<Zval>` is not implemented for all types AFAIK, what about `T: FromZval<'a>` taking `&'a self` instead?

Or if you want to drop `self` then `for<'a> T: FromZval<'a>` and keep the function parameter as `self`. The `for<'a>` just makes it valid to have `T` as `String` and not `&str` (as it would reference the zval after dropping).

src/php/globals.rs
davidcole1340: This is fine for now since it's not getting used anywhere, but eventually should be contained in a `ZBox`.

src/php/globals.rs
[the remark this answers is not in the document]
davidcole1340: I think we can look at synchronisation later. Having some is probably better than none, so if we can ensure it on the Rust side then I'd say that's better than nothing. But that can be done in a different PR.

src/php/types/object.rs
```diff
@@ -157,6 +159,65 @@ impl ZendObject {
...
+/// let obj: ZendObject = ...;
+/// let repr: String = obj.extract();
+/// let props: HashMap = obj.extract();
+/// ```
+///
+/// Should be functionally equivalent to casting an object to another compatable type.
+pub trait FromZendObject: Sized {
+    /// Extracts `Self` from the source `ZendObject`.
+    fn extract(obj: &ZendObject) -> Result<Self>;
+}
```
davidcole1340, line 185:
```suggestion
pub trait FromZendObject<'a>: Sized {
    /// Extracts `Self` from the source `ZendObject`.
    fn extract(obj: &'a ZendObject) -> Result<Self>;
}
```

Could we add a lifetime on the `ZendObject` reference? It will allow the trait to be implemented on types which will contain a reference to something inside the object (e.g. a property holding a string, returning a struct holding a `&str`).

src/php/types/object.rs
```diff
@@ -157,6 +159,65 @@ impl ZendObject {
...
+
+impl FromZendObject for String {
```
davidcole1340, line 187:
```suggestion
impl FromZendObject<'_> for String {
```

As above.

src/php/types/object.rs
```diff
@@ -157,6 +159,65 @@ impl ZendObject {
...
+/// ```rust,ignore
+/// let obj: ZendObject = ...;
+/// let repr: String = obj.extract();
+/// let props: HashMap = obj.extract();
+/// ```
+///
+/// Should be functionally equivalent to casting an object to another compatable type.
+pub trait FromZendObject: Sized {
+    /// Extracts `Self` from the source `ZendObject`.
+    fn extract(obj: &ZendObject) -> Result<Self>;
```
davidcole1340, line 184:
```suggestion
    fn from_zend_obj(obj: &ZendObject) -> Result<Self>;
```

Perhaps different name?

vodik: Ah, went ahead and changed it and then saw this. Agreed. Wondered why the hell I even called it `extract` in the first place.

PR #76, opened 2021-09-20
Refactor `ZendHashTable` and `Zval`

davidcole1340:
- `ZendHashTable` has been split into `&HashTable` and `OwnedHashTable`. `&HashTable` is to `OwnedHashTable` as `&str` is to `String`.
- The `Drop` implementation on `Zval` will properly drop the contents of the zval.

### DIFF
--- a/src/php/types/array.rs
+++ b/src/php/types/array.rs
@@ -6,6 +6,7 @@ use std::{
     convert::{TryFrom, TryInto},
     ffi::CString,
     fmt::Debug,
+    mem::ManuallyDrop,
     ops::{Deref, DerefMut},
     ptr::NonNull,
     u64,
@@ -254,7 +255,6 @@ impl<'a> Iterator for Iter<'a> {
 /// function is used.
 pub struct OwnedHashTable {
     ptr: NonNull<HashTable>,
-    drop: bool,
 }
 
 impl OwnedHashTable {
@@ -294,14 +294,13 @@ impl OwnedHashTable {
     pub unsafe fn from_ptr(ptr: *mut HashTable) -> Self {
         Self {
             ptr: NonNull::new(ptr).expect("Invalid hashtable pointer given"),
-            drop: true,
         }
     }
 
     /// Returns the inner pointer to the hashtable, without destroying the
-    pub fn into_inner(mut self) -> *mut HashTable {
-        self.drop = false;
-        self.ptr.as_ptr()
+    pub fn into_inner(self) -> *mut HashTable {
+        let this = ManuallyDrop::new(self);
+        this.ptr.as_ptr()
     }
 }
 
--- a/src/php/types/array.rs
+++ b/src/php/types/array.rs
@@ -378,7 +378,7 @@ where
         for (idx, key, val) in zht.into_iter() {
             hm.insert(
                 key.unwrap_or_else(|| idx.to_string()),
-                V::from_zval(val).ok_or(Error::ZvalConversion(val.get_type()?))?,
+                V::from_zval(val).ok_or(Error::ZvalConversion(val.get_type()))?,
             );
         }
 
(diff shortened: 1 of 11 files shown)


src/php/types/array.rs
```diff
@@ -277,202 +198,299 @@ impl<'a> Debug for ZendHashTable<'a> {
...
-    fn clone(&self) -> Self {
-        // SAFETY: If this fails then `emalloc` failed - we are doomed anyway?
-        // `from_ptr()` checks if the ptr is null.
+/// A container used to 'own' a Zend hashtable. Dereferences to a reference to [`HashTable`].
+///
+/// When this struct is dropped, it will also destroy the internal hashtable, unless the `into_raw`
+/// function is used.
+pub struct OwnedHashTable {
+    ptr: NonNull<HashTable>,
+    drop: bool,
```
vodik, line 257: Unless I'm missing something, I'm not seeing it read anywhere,

src/php/types/array.rs
```diff
@@ -277,202 +198,299 @@ impl<'a> Debug for ZendHashTable<'a> {
...
+            ptr: NonNull::new(ptr).expect("Invalid hashtable pointer given"),
+            drop: true,
+        }
+    }
+
+    /// Returns the inner pointer to the hashtable, without destroying the
+    pub fn into_inner(mut self) -> *mut HashTable {
+        self.drop = false;
+        self.ptr.as_ptr()
+    }
```
vodik, line 305: That said, is `self.drop` just intended for bookkeeping to prevent us from freeing memory moved out of the type? You should be able to do the same thing with `std::mem::forget`:

```suggestion
    pub fn into_inner(mut self) -> *mut HashTable {
        let ptr = self.ptr;
        std::mem::forget(self);
        ptr
    }
```

Or use `ManuallyDrop` under the hood.

vodik: And if you don't need drop, then `OwnedHashTable` could be defined as a newtype wrapper:
```rust
pub struct OwnedHashTable(NonNull<HashTable>)
```

src/php/types/array.rs
```diff
@@ -277,202 +198,299 @@ impl<'a> Debug for ZendHashTable<'a> {
...
 
-/// Implementation for converting a reference to `ZendHashTable` into a `Vec` of given type.
-/// Will return an error type if one of the values inside the array cannot be converted into
-/// a type `T`.
-impl<'a, V> TryFrom<ZendHashTable<'a>> for Vec<V>
+///////////////////////////////////////////
+//// Vec
+///////////////////////////////////////////
+
+impl<T> TryFrom<&HashTable> for Vec<T>
```
vodik, line 438: Could we get array with `FromIterator` for this? It would be a bit more powerful as you can collect into either `Result<Vec<T>>` or Vec<Result<T>>`:
```rust
let vec: Result<Vec<u32>> = hashmap.iter()
    .map(|(_, _, val)| val.extract()
    .collect();
```

vodik (approved):
I think it looks reasonable to me. Just a few thoughts.

src/php/types/array.rs
vodik: Actually, its probably better to just provide more iterators:
```rust
let vec: Result<Vec<u32>> = hashmap.iter_values()
    .map(Zval::extract)
    .collect();
```
And this probably is more flexible as well, as PHP datastructures are not guaranteed to be heterogeneous in the first place.

src/php/types/array.rs
davidcole1340: Yeah, I had changed this on the `zend-str` branch after reading the `CStr` documentation

src/php/types/array.rs
davidcole1340: Yeah, iterator is probably better, I only added the impls for ease of use and this way preallocates using `with_capacity`, although now thinking about it the `FromIterator` implementations probably use the `count` function on `Iterator`.

Good idea about the `values()` iterator, will add  it now.

vodik: My understand is it may if there's a size hint provided.

But it's probably best to leave it to the target collection to decide as well.

davidcole1340: Do you think these implementations should be removed in favour of iterators? I will also implement `ExactSizeIterator` and `DoubleEndedIterator` for `HashTable`

PR #77, opened 2021-09-21
Refactor Zend strings

davidcole1340:
Zend strings now come in two forms - the borrowed `&ZendStr` and the owned `ZendString`, which can be compared to `&str` and `String`.

### DIFF
--- a/example/skel/test.php
+++ b/example/skel/test.php
@@ -1,6 +1,10 @@
 <?php
 
+<<<<<<< HEAD
 test_zval(['hello' => 'world']);
+=======
+test_zval();
+>>>>>>> 712aded56 (Call zval destructor when changing zval type and dropping)
 //include 'vendor/autoload.php';
 
 //$ext = new ReflectionExtension('skel');
--- a/example/skel/test.php
+++ b/example/skel/test.php
@@ -1,10 +1,14 @@
 <?php
 
 <<<<<<< HEAD
+<<<<<<< HEAD
 test_zval(['hello' => 'world']);
 =======
 test_zval();
 >>>>>>> 712aded56 (Call zval destructor when changing zval type and dropping)
+=======
+test_zval(['hello' => 'world']);
+>>>>>>> 3646da741 (Remove `ZendHashTable` wrapper)
 //include 'vendor/autoload.php';
 
 //$ext = new ReflectionExtension('skel');
--- a/example/skel/test.php
+++ b/example/skel/test.php
@@ -2,6 +2,7 @@
 
 <<<<<<< HEAD
 <<<<<<< HEAD
+<<<<<<< HEAD
 test_zval(['hello' => 'world']);
 =======
 test_zval();
@@ -9,6 +10,9 @@ test_zval();
 =======
 test_zval(['hello' => 'world']);
 >>>>>>> 3646da741 (Remove `ZendHashTable` wrapper)
+=======
+test_zval([]);
+>>>>>>> 7ba9ee9bf (Refactor `ZendString` into a borrowed and owned variant)
 //include 'vendor/autoload.php';
 
 //$ext = new ReflectionExtension('skel');
--- a/src/php/class.rs
+++ b/src/php/class.rs
@@ -38,15 +38,11 @@ impl ClassEntry {
     /// not be found or the class table has not been initialized.
     pub fn try_find(name: &str) -> Option<&'static Self> {
         ExecutorGlobals::get().class_table()?;
-        let name = ZendString::new(name, false).ok()?;
+        let mut name = ZendString::new(name, false).ok()?;
 
         unsafe {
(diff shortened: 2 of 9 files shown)


example/skel/test.php
```diff
@@ -1,6 +1,18 @@
 <?php
 
+<<<<<<< HEAD
```
davidcole1340, line 3: Forgot to rebase this properly, but I'm going to delete `example/skel` soon as it's just my testbench at this point

src/php/types/string.rs
```diff
@@ -1,145 +1,315 @@
...
-    /// Creates a new [`ZendString`] wrapper from a raw pointer to a [`zend_string`].
+    /// * `str` - String content.
+    /// * `persistent` - Whether the string should persist through the request boundary.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the function was unable to allocate memory for the Zend string.
+    pub fn from_c_str(str: &CStr, persistent: bool) -> Self {
+        let ptr = unsafe {
+            ext_php_rs_zend_string_init(str.as_ptr(), str.to_bytes().len() as _, persistent)
```
vodik, line 164: For the sake of producing a safe and correct wrapper around this, the `persistent` field probably should be hoisted into the type. It is not safe to provide request-arena allocated strings, for example, to a function that expects persisted ones - while the opposite is likely fine. We have covariance.

I don't think this can be easily expressed via lifetimes, but should be expressible with type variables:
```rust
struct ZendStringBase<Alloc>;

pub type ZendString = ZendStringBase<Request>;

pub mod persistent {
    pub type ZendString = ZendStringBase<Persistent>;
}

Comment (vodik) [reply src/php/types/string.rs:164]: As I understand it, the above should be perfectly fine so long as we're content with only using rust to extend PHP.

However, I'm not super familiar with how request allocated memory interacts with embedded PHP. I assume the lifetime of a request might be in the control of the hosting language.

If so, that opens the door to another safety issue: use after free after the request arena is deallocated. So another option might be to make `ZendString::new` _always_ persistent and instead add some sort of token that represents the current request into the API - even if its largely an artificial abstraction.

This request object then must used to allocate per-request data and can be used to coordinate lifetimes

Comment (vodik) [reply src/php/types/string.rs:164]: The detail I'm still not sure about, I've been digging through php-src to figure this out, is where does the emalloc heap get cleared.

File: src/php/class.rs [hunk 3]
```diff
@@ -41,8 +41,12 @@ impl ClassEntry {
         let name = ZendString::new(name, false).ok()?;
 
         unsafe {
-            crate::bindings::zend_lookup_class_ex(name.borrow_ptr(), std::ptr::null_mut(), 0)
-                .as_ref()
+            crate::bindings::zend_lookup_class_ex(
+                name.as_zend_str() as *const _ as *mut _,
```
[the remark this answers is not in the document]
davidcole1340: Yes, I was worried about users using it to mutate the string but with no setters that shouldn't be a problem (although bindgen generates structs with `pub` fields, but regardless the function could be `pub(crate)`).

src/php/types/string.rs
davidcole1340: Forcing all owned Zend strings to be persistent might be the best way I think, calling `zend_string_release` once their GC cycle is zero still causes them to be deallocated, however we lose the memory leak checks of the ZMM.

I'm not sure how we could account for the user potentially keeping a request-bound `ZendString` past the request boundary, leaving a dangling pointer inside of `ZendString`. Maybe just forcing all strings to be persistent would be the safest option?

vodik: > I'm not sure how we could account for the user potentially keeping a request-bound ZendString past the request boundary, leaving a dangling pointer inside of ZendString

Explicitly creating a token type that wraps the request is probably the way to go: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c431c93f58529be80decb279e61589bc

In the example, even through `TEST` is static, if I grab it through `fake_alloc` I cannot get a borrow that outlives the Request token type.

Alt with a closure: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c431c93f58529be80decb279e61589bc

davidcole1340: How would we represent the lifetime of the request though? A `Request` object could either live for the lifetime of the `extern "C"` wrapper function or `'static` (or maybe I am missing something)?

The `ZendString` API _is_ technically an 'internal' API, so perhaps it could be deemed unsafe (or default to persistent with an unsafe request-bound option)?

davidcole1340: So it seems that you aren't supposed to put persistent strings in zvals... When calling `zend_ptr_dtor()` (from the `Drop` impl or when called from PHP) it calls `zend_string_destroy` which only attempts to deallocate using the Zend memory manager.

davidcole1340:
Going to merge this for now and can revisit later.

Issue #75, opened 2021-09-20
Add `FromZval` derive macro


davidcole1340:
Ability to derive `FromZval` (and potentially `IntoZval`) using a derive macro. Could be used in two situations:

- Enums.  Data type would be mixed, and would act as a mixed type parameter. Example:
```rust
#[derive(FromZval)]
pub enum IntOrBool {
    Integer(u32),
    Boolean(bool),
}

// translates to

impl FromZval<'_> for IntOrBool {
    const TYPE: DataType = DataType::Mixed;
    fn from_zval(zval: &'_ Zval) -> Option<Self> {
        if let Some(integer) = u32::from_zval(zval) {
            Some(integer)
        } else if let Some(boolean) = bool::from_zval(zval) {
            Some(boolean)
        } else {
            None
        }
    }
}
```

- Structs. Data type would be object, and the struct fields would be read from the objects properties. Primarily for passing `stdClass` into Rust. Example:

```rust
#[derive(FromZval)]
pub struct ExampleStdClass {
    a: i32,
    b: String,
    c: bool,
}

// translates to

impl FromZval<'_> for ExampleStdClass {
    const TYPE: DataType = DataType::Object(None);

    fn from_zval(zval: &'_ Zval) -> Option<Self> {
        let obj = zval.object()?;

        Some(Self {
            a: obj.get_property("a").ok()?,
            b: obj.get_property("b").ok()?,
            c: obj.get_property("c").ok()?,
        })
    }
}
```

vodik:
I like it. It should be paired with a `IntoZval` as well.

Something worth keeping an eye on is [PHP 8.1 enums](https://stitcher.io/blog/php-enums). I don't think this should conflict through: a reasonable approach is probably to either add a new `#[php_enum]` attribute (or make `php_class` smarter - idk how its implemented under the hood) and make sure that its mutually exclusive from these derive macros.

PR #78, opened 2021-09-23
Anonymous class `FromZval` derive macro

davidcole1340:
Progress towards #75 

Added the `#[derive(FromZval)]` macro. Can be used on enums and structs.

- Enums: Treats the object as a tagged union mixed parameter.
- Struct: Treats the object as an anonymous class, filling the fields
  from an objects properties.

Both support generics and lifetimes so `&'a str` is valid. Next step in a
seperate PR is the corresponding `IntoZval` macro, which converts the
class into a `Zval`.


ext-php-rs-derive/src/lib.rs
```diff
@@ -125,3 +127,15 @@ pub fn php_extern(_: TokenStream, input: TokenStream) -> TokenStream {
     }
     .into()
 }
+
+// TODO(david): potentially add option to ignore field in exchange for `Default::default()`?
```
davidcole1340, line 131: Is this needed? I'm not sure if I see the purpose of having a field with a default value but it seems simple enough to add.

davidcole1340:
Also started work on the inverse `#[derive(IntoZval)]` macro: https://github.com/davidcole1340/ext-php-rs/compare/from_zval_stdclass...into_zval_stdclass

ext-php-rs-derive/src/from_zval.rs
```diff
@@ -0,0 +1,161 @@
...
+            const TYPE: ::ext_php_rs::php::enums::DataType = ::ext_php_rs::php::enums::DataType::Object(::std::option::Option::None);
+
+            fn from_zval(zval: &'_zval ::ext_php_rs::php::types::zval::Zval) -> ::std::option::Option<Self> {
+                let obj = zval.object()?;
+
+                Some(Self {
+                    #(#fields)*
+                })
+            }
+        }
```
davidcole1340, line 102: Thinking about this more, perhaps we should introduce a new trait, `FromZendObject`, as well as the corresponding `IntoZendObject`?

vodik:
I really like this, but I think there are some interesting semantics to explore. I've been thinking about this for a few days and trying to get all my thoughts down, so apologies of this is somewhat unstructured.

My first through was that if I was to see the rust code that this feature would example, for example something like this:
```rust
#[derive(FromZVal)]
struct Args {
    name: String
}

#[php_function]
fn printer(args: Args) {
    println!("{}", args.name)
}
```

I might intuit that the following PHP code to work:
```php
printer(['name' => 'Simon'])
```

Rather than this:
```php
class Args {
    public function __construct(public string $name) {}
}

printer(new Args)
```

However *both* would be useful, and should work given PHP's own casting rules. But the catch is its part of PHP's *casting* rules. To write a PHP version than can take either I have to explicitly cast:

```php
function printer(object|array $args) {
    $args = (object)$args
    echo "{$args->name}\n";
}
```
or
```php
function printer(object|array $args) {
    $args = (array)$args
    echo "{$args['name']}\n";
}
```

I cannot rely on implicit conversion here.

# `FromZVal` only unpacks into PHP types *or* enums representing union of types

Keep it as close to php type annotations as possible:

| PHP Type | Rust Type  |
| -------- | ---------- |
| array    | HashMap    |
| callable | Closure    |
| bool     | bool       |
| float    | f32        |
| int      | u32/u64    |
| string   | String     |
| iterable | ?          |
| object   | ZendObject |
| mixed    | ZVal       |

Then nullable types turn into option types: `?string` is represented as `Option<String>`

And union types are represented as enums: `string|object` would be represented as:
```rust
#[derive(FromZVal)]
enum CustomType {
    String(String),
    Object(ZendObject)
}
```

# `FromZVal` should round trip by default

The nice part of keeping it simple like this is that all types can predictably round trip. Any type that can have `FromZVal` could also have a valid and expected `IntoZVal` (maybe we don't even need an explicit `IntoZVal`).

# Represent casts with a different API, make them explicit

I like this functionality in this PR, but I'd suggest making is an explicit function call with a district trait/derive name. For example:

```rust
// Define this so function can take either array or object.
// Represents array|object
#[derive(FromZVal)]
enum Args {
    Object(ZendObject),
    Array(HashMap),
}

// Tentative name, bikeshedable
#[derive(CastFromZVal)]
struct Props {
    name: String
}

#[php_function]
fn printer(args: Args) {
    // Unpack either the object or the array into `Props`
    let props = php::cast(&args);
    println!("{}", props.name)
}
```

# Current `FromZVal` attaches behaviour to the wrong layer

One other note I think is worth considering with the current design is that this `FromZVal` implementation currently first unpacks the `ZVal` to an object (`ZendObject`) *and then* to into the struct. This means if I just have an object, I can't unpack it without turning it into a `ZVal` first.

My other PR introduces `FromZendObject` trait and this is probably better organized under that trait. Within the current design, I'd advocate its more correct for from `FromZVal` to delegate to `FromZendObject` rather than having it all in one. Then you could get `ZVal` to `String` (via `__toString`) for free as well.

# declare_strict

We might also want to consider strict vs not strict conversion rules in `FromZVal`. We might want to support both strict and non-strict versions of `FromZVal`:

```rust
let zval = 2.3f.into_zval().unwrap();

let a: Result<u64> = zval.extract_loosly();  // Holds Ok(2)
let b: Result<u64> = zval.extract();         // Holds Err(...)
let c: Result<String> = php::cast(&zval);    // Holds "2.3"
```

On that note, this is something I just learned, that `__toString` allows classes to be implicitly cast to strings. The following works unless its strict mode:

```php
class Args {
    public function __construct(public string $name) {}
    
    public function __toString() { return $this->name; }
}

function printer(string $msg) {
    echo "$msg\n";
}

printer('Hello World');

// Not a type error
printer(new Args('Hello World'))
```

So to reiterate the above:

```rust
let obj: ZVal = ...;

let a: Result<String> = obj.extract_loosly();  // Calls `__toString`
let b: Result<String> = obj.extract();         // Err(...)
let c: Result<String> = php::cast(&obj);       // Calls `__toString`
```

ext-php-rs-derive/src/function.rs
```diff
@@ -217,7 +217,8 @@ impl Arg {
     pub fn new(name: &str, ty: &str, nullable: bool, default: Option<String>) -> Self {
         Self {
             name: name.to_string(),
-            ty: Regex::new(r"'[A-Za-z]+")
+            // remove lifetimes - wish there was a better way :(
+            ty: Regex::new(r"'[A-Za-z ]+[,]?")
```
vodik, line 221: Might not hurt to ask the syn/quote people.

ext-php-rs-derive/src/function.rs
```diff
@@ -361,7 +362,7 @@ impl Function {
             })
             .collect::<Vec<_>>();
         let output = self.output.as_ref().map(|(ty, nullable)| {
-            let ty: Type = syn::parse_str(ty).unwrap();
+            let ty: Type = syn::parse_str(ty).expect("failed to parse ty: function.rs:365");
```
vodik, line 365: Isn't the filename/line number be part of the panic?

vodik (approved):
Merge it in. The finer details of how casting and coersion work are better dealt with this in tree rather than out of tree.

I like it.

ext-php-rs-derive/src/function.rs
davidcole1340: Yep 😆

PR #83, opened 2021-09-28
Add class constructors

davidcole1340:
Constructor methods can now be specified. They are required if
you want to construct the class from PHP. This replaces the `Default`
implementation requirement on class objects.

Constructors can return `Self` or `Result<Self, E>` where `E: Into<PhpException>`.

### DIFF
--- a/src/php/types/object.rs
+++ b/src/php/types/object.rs
@@ -28,8 +28,10 @@ use crate::{
     php::{
         class::ClassEntry,
         enums::DataType,
-        exceptions::PhpResult,
+        exceptions::{PhpException, PhpResult},
+        execution_data::ExecutionData,
         flags::ZvalTypeFlags,
+        function::FunctionBuilder,
         globals::ExecutorGlobals,
         types::{array::OwnedHashTable, string::ZendString},
     },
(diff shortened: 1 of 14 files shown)


src/php/types/object.rs
```diff
@@ -352,31 +417,43 @@ impl<'a, T: RegisteredClass> FromZval<'a> for &'a T {
...
 
-        Some(unsafe { &*cobj.obj.as_mut_ptr() })
+        Some(unsafe { &*cobj.obj.as_ptr() })
     }
 }
 
-impl<'a, T: RegisteredClass> FromZval<'a> for &'a mut T {
-    const TYPE: DataType = DataType::Object(Some(T::CLASS_NAME));
+// TODO(david): mutable FromZval?
+// impl<'a, T: RegisteredClass> FromZval<'a> for &'a mut T {
```
davidcole1340, line 427: Because `Zval::object()` was changed to return an immutable reference, `FromZval` can no longer be implemented for `&mut T`. I think we will need to add a new trait `FromZvalMut`, and then any type which implements `FromZval` can implement `FromZvalMut`.

davidcole1340: Perhaps this should be done in a later PR as it will require keeping a mutable `Zval` inside `Arg`

vodik (approved):

PR #92, opened 2021-10-01
Add zval type coersion

davidcole1340:
`Zval::string()` will no longer attempt to convert a long into a string. Instead, this is done through the new function on the `FromZval` trait: `from_zval_coerce`, which broadly follows PHP's [type juggling rules].

By default, downstream types do not have to implement this function, as its default is to call `from_zval` on the same type.

[type juggling rules]: https://www.php.net/manual/en/language.types.type-juggling.php

### DIFF
--- a/src/php/types/zval.rs
+++ b/src/php/types/zval.rs
@@ -844,8 +844,8 @@ impl FromZval<'_> for bool {
         zval.bool()
             .or_else(|| zval.long().map(|l| l != 0))
             .or_else(|| zval.double().map(|d| d != 0.0))
-            .or_else(|| zval.str().map(|s| !(s.len() == 0 || s == "0") || s == "1"))
-            .or_else(|| zval.array().map(|arr| arr.len() != 0))
+            .or_else(|| zval.str().map(|s| !(s.is_empty() || s == "0") || s == "1"))
+            .or_else(|| zval.array().map(|arr| !arr.is_empty()))
             .or_else(|| if zval.is_null() { Some(false) } else { None })
     }
 }
--- a/src/php/types/zval.rs
+++ b/src/php/types/zval.rs
@@ -76,14 +76,11 @@ impl Zval {
         if self.is_double() {
             Some(unsafe { self.value.dval })
         } else {
-            self.long().map(|x| x as f64)
+            None
         }
     }
 
     /// Returns the value of the zval as a zend string, if it is a string.
-    ///
-    /// Note that this functions output will not be the same as [`string()`](#method.string), as
-    /// this function does not attempt to convert other types into a [`String`].
     pub fn zend_str(&self) -> Option<&ZendStr> {
         if self.is_string() {
             unsafe { self.value.str_.as_ref() }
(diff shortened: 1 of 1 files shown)


src/php/types/zval.rs
```diff
@@ -733,51 +732,122 @@ where
...
+            .or_else(|| zval.long().map(|l| l.to_string()))
+            .or_else(|| zval.double().map(|d| d.to_string()))
+            .or_else(|| zval.bool().map(|b| if b { "1" } else { "" }.to_string()))
+            .or_else(|| {
+                if zval.is_null() {
+                    Some("".to_string())
+                } else {
+                    None
+                }
+            })
```
vodik, line 806: Also only a cast, not a coercion.

src/php/types/zval.rs
```diff
@@ -733,51 +732,122 @@ where
...
+    const TYPE: DataType = DataType::Bool;
+
+    fn from_zval(zval: &Zval) -> Option<Self> {
+        zval.bool()
+    }
+
+    fn from_zval_coerce(zval: &Zval) -> Option<Self> {
+        // https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting
+        zval.bool()
+            .or_else(|| zval.long().map(|l| l > 0))
```
vodik, line 845: Oh, these are wrong. Its the same rules as C: zero is false, everything else is true.
```suggestion
            .or_else(|| zval.long().map(|l| l != 0))
```

src/php/types/zval.rs
```diff
@@ -733,51 +732,122 @@ where
...
+            .or_else(|| zval.double().map(|d| d > 0.0))
+            .or_else(|| zval.str().map(|s| !(s.is_empty() || s == "0") || s == "1"))
```
vodik, line 847: Any other string is considered true.
```suggestion
            .or_else(|| zval.str().map(|s| !s.is_empty() && s != "0"))
```

src/php/types/zval.rs
```diff
@@ -733,51 +732,122 @@ where
...
+            .or_else(|| zval.array().map(|arr| !arr.is_empty()))
+            .or_else(|| if zval.is_null() { Some(false) } else { None })
```
vodik, line 849: In casting, not coercion.

vodik (commented):
I like it, but looks like there's been some lingering bugs here with incorrect login and some overlap in casting and coercion rules.

Its probably best we don't do coersions from null. Just like how php will now insist nulliblity is included in type annotations, we should as require an explicit use of `Option`. Or an explicit cast. (It also makes me a bit more confident in my assertion that casting should be a free function).

danog:
Personally, not a huge fan of this, closing this for now unless someone wants to pick up the implementation and fix outstanding issues.

I really don't like PHP's type coercion behavior, and in fact, I believe that the closer this extension is to `declare(strict_types=1)` behavior, the better.

https://github.com/davidcole1340/ext-php-rs/pull/298 removes existing type coercion without adding replacement logic.

PR #94, opened 2021-10-03
Add `ZBox<T>` to replace owned variants

davidcole1340:
Currently types which cannot be owned directly are owned through wrapper types,  such as `ZendString -> ZendStr`, `OwnedHashTable -> HashTable`, `ClassObject -> ZendObject`, since they are allocated using the Zend memory manager and cannot be stored in a box.

This PR introduces a new type `ZBox<T>` and a new trait `ZBoxable`. A type allocated using the Zend memory manager will implement `ZBoxable`, and can then be stored inside a `ZBox<T>`.

There are no 'safe' constructor methods on `ZBox`. These are to be implemented on the boxed types. For example, `ZendStr` has a function `from_c_str` which returns a `ZBox<ZendStr>`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
         env:
           LIBCLANG_PATH: ${{ runner.temp }}/llvm-${{ matrix.llvm }}/lib
           EXT_PHP_RS_TEST:
-        run: cargo build --release --features alloc,closure
+        run: cargo build --release --all-features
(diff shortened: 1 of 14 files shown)


davidcole1340:
At the moment I've only implemented `ZBoxable` for `ZendStr` but will do after a quick review. Can be implemented on the following types:

- `HashTable`
- `ZendObject`
- `ZendClassObject`

PR #101, opened 2021-10-07
Refactor module paths

### LINKED ISSUE #91: Reorganise modules
Module paths are currently long and convoluted. Can easily be shortened (drop `::php`, re-export types from `::types` instead of exporting modules).

davidcole1340:
Resolves #91


vodik:
Late to the party, but that's a big improvement!

PR #121, opened 2022-01-13
Add integration tests

TorstenDittmann:
Implemented a very basic test suite for all Datatypes that can be passed as an argument or returned.

I am not sure what kind of conventions you would like to follow here, so I kept the PHP files very stupid - same with the assertion texts. So we would need to decide on a proper way of how to pass the results from PHP to Rust.

Not a 100% happy how the PHP code is executed using a hardcoded path to the compiled extension. Nontheless, I think that is a good start for the first iteration 👍🏻


tests/src/integration/array.rs
```diff
@@ -0,0 +1,4 @@
+#[test]
+fn binary_works() {
+    assert!(crate::integration::run_php("array.php"));
```
vodik, line 3: > Not a 100% happy how the PHP code is executed using a hardcoded path to the compiled extension. Nontheless, I think that is a good start for the first iteration

And idea that might work is to put together a simple macro and embed the php snippet directly into this file:

```rust
#[test]
fn binary_works() {
    php_test!(r#"
// Tests sequential arrays
$array = test_array(['a', 'b', 'c']);

assert(is_array($array));
assert(count($array) === 3);
assert(in_array('a', $array));
assert(in_array('b', $array));
assert(in_array('c', $array));

// Tests associative arrays
$assoc = test_array_assoc([
    'a' => '1',
    'b' => '2',
    'c' => '3'
]);

assert(array_key_exists('a', $assoc));
assert(array_key_exists('b', $assoc));
assert(array_key_exists('c', $assoc));
assert(in_array('1', $assoc));
assert(in_array('2', $assoc));
assert(in_array('3', $assoc));
    "#);
}
```
For the initial version, we could either automatically write out a file an execute it or - I believe this should work - pipe the input into php without touching the filesystem.

I will resume working on my execution support, after which this could mature into something entirely in process.

vodik (commented):
Will do a deeper dive later tonight, this thought just crossed my mind.

vodik (approved):
Looks like a solid start to me :+1:

vodik:
@TorstenDittmann how are you feeling about that state of this PR? I may be able to help finish it if you're short on time.

joelwurtz:
What's the status on this ? I feel like such test would be very helpful when developing this library

ptondereau:
> What's the status on this ? I feel like such test would be very helpful when developing this library
It's the most critical feature that I want, but this PR is abandoned.
IMHO, I would try to reproduce what php core does with `phpt`

TorstenDittmann:
Yeah, it kinda got abandoned due to many reason.

If there is interest I can spend a weekend to probably create the PR again from scratch 👀

joelwurtz:
> IMHO, I would try to reproduce what php core does with phpt

That would be nice, it could even be used by people creating extension with this lib to test there code.

I think this would be possible if there was a possibility to compile php with code from this library under a new binary, which would also allow to create binary in rust that can execute php code (like a new sapi ?)

ptondereau:
> That would be nice, it could even be used by people creating extension with this lib to test there code.
> 
> I think this would be possible if there was a possibility to compile php with code from this library under a new binary, which would also allow to create binary in rust that can execute php code (like a new sapi ?)

There is this experimental module here https://github.com/php/php-src/tree/master/sapi/embed to produce a single binary with the Zend engine entry point. I did a small PoC with it and Tauri to reproduce what [NativePHP](https://nativephp.com/) does: it works well.
Instead of compiling into a whole extension, we could execute the tested code with it

joelwurtz:
Can you share your PoC with Tauri, i would be interested in how you did that

joelwurtz:
I made an example on https://github.com/davidcole1340/ext-php-rs/pull/270 

It works great, it even allow to put test directly in the wanted file, i will try to do a wrapper so we can execute code directly in it, thanks for the input @ptondereau about embed this is certainly the best way to go for that

PR #128, opened 2022-03-13
Preliminary Windows support

davidcole1340:
Still some work to do but will use this to track progress.

Some caveats:
- Rust nightly is required. PHP uses `__vectorcall` on some functions which is supported in Rust via the `abi_vectorcall` nightly feature.
- Bindgen `__vectorcall` support has not been upstreamed yet and therefore requires using my fork.
- Only PHP installations from windows.php.net are supported at this point in time.
- You need to match MSVC (Microsoft Visual C++) versions with PHP. The extension must be linked with a MSVC version the same or earlier than the version of MSVC PHP was linked with. Visual Studio 2022 is not supported.
- I don't think tests are passing at the moment on Windows.
- If you manually define function handlers (i.e. ones that take Zend exceution data) you need to change to using the `zend_fastcall!` macro. Function handlers are `__vectorcall` on Windows and the macro will swap the ABI based on the target triple. Example:
```rs
// Before
pub extern "C" fn test_handler(ex: &mut ExecutionData, retval: &mut Zval) {
  // ...
}

// After
use ext_php_rs::zend_fastcall;

zend_fastcall! {
  pub extern fn test_handler(ex: &mut ExecutionData, retval: &mut Zval) {
    // ...
  }
}
```


SQKo:
wintendo user here approves

davidcole1340:
Some updates:
- CI now passing on Windows
- Bindgen patches have been upstreamed but not released: https://github.com/rust-lang/rust-bindgen/pull/2177 
- Now using rust-lld linker (LLVM linker) so don't need to worry about matching linker versions (MSVC is still required to compile C wrapper)

reeperbahnause:
Hi, I just tried to build the hello_world example on my local Windows but i still get the error regarding the missing php-config. Do you have a hint for what I'm missing?

SQKo:
> Hi, I just tried to build the hello_world example on my local Windows but i still get the error regarding the missing php-config. Do you have a hint for what I'm missing?

Hey, It's working for me on windows.
Are you using the latest `master` branch? Try it

Otherwise please post the details of the errors.

Issue #134, opened 2022-05-04
PHP INI file can be fully truncated or contents duplicated when using `cargo-php`


roborourke:
I noticed then when using `cargo-php remove` the INI file contents are completely truncated.

Additionally when using `cargo-php install` the file contents are duplicated leading to the `extension=` lines doubling up and PHP throwing an error due to trying to load `libv8js.soextension` or similar.

PR #135, opened 2022-05-04
Allow passing `--yes` parameter to bypass prompts

roborourke:
Makes this tool usable in automated builds such as Docker containers.

Addresses https://github.com/davidcole1340/ext-php-rs/issues/133

PR #136, opened 2022-05-04
Fix ini file duplication and truncation when using `cargo-php` command

roborourke:
Addresses https://github.com/davidcole1340/ext-php-rs/issues/134

Also prevents the extension line from being added twice on subsequent calls.


roborourke:
Whoops, some unintended changes from #135 made it into this one

davidcole1340:
Thank you!

PR #146, opened 2022-07-25
Add php-scrypt as a example project

PineappleIOnic:
Hey There 👋 

We at Appwrite recently created an extension using this library to fix an issue we had with the standard C-based php-scrypt by rewriting it in rust. Thanks to this library being so easy to use, we could get a working version in less than a day! 

We would love it if you could add it as an example project in this repository to give more examples of how to use this library.

Thank you for reviewing this pull request,
Bradley Schofield

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,8 @@ Check out one of the example projects:
 - [opus-php](https://github.com/davidcole1340/opus-php) - Audio encoder for the
   Opus codec in PHP.
 - [tomlrs-php](https://github.com/jphenow/tomlrs-php) - TOML data format parser.
+- [php-scrypt](https://github.com/appwrite/php-scrypt) - PHP wrapper for the
+  scrypt password hashing algorithm.
 
 ## Contributions
 


ptondereau (approved):
Thank you for your example!
I see that you don't have any PHP stub in your library.
You can use this mini tutorial https://github.com/ptondereau/ext-redlock-php#generating-php-stubs (because the `cargo stubs` command is unstable for the moment)

PineappleIOnic:
@ptondereau Thank you for reviewing our example, I was wondering where we should place our stubs within the repo once we have generated them? Is there a standard for this or should we do add it to our build steps?

ptondereau:
> @ptondereau Thank you for reviewing our example, I was wondering where we should place our stubs within the repo once we have generated them? Is there a standard for this? or should we add it to our build steps?

IMHO, it's up to you: I would go to exposing the stubs with a dedicated Packagist entry. In the ext's repository, add all stubs to a composer.json with a name in a manner that your user just need to `composer install vendor/ext-stubs`

PR #155, opened 2022-09-21
Support marking classes as interfaces

joehoyle:
This allows passing flags as part of `#[php_class(flags=Interface]` etc, which allows one to mark a class as being an interface.

When a class is an interface, it also shouldn't get a constructor created for it.

### DIFF
--- a/crates/macros/src/class.rs
+++ b/crates/macros/src/class.rs
@@ -22,6 +22,7 @@ pub struct Class {
     /// A function name called when creating the class entry. Given an instance
     /// of `ClassBuilder` and must return it.
     pub modifier: Option<String>,
+    pub flags: Option<String>
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


striezel:
@joehoyle: There are some problems with `rustfmt`:

```
 /home/runner/.cargo/bin/cargo fmt --all -- --check
Diff in /home/runner/work/ext-php-rs/ext-php-rs/crates/macros/src/class.rs at line 22:
     /// A function name called when creating the class entry. Given an instance
     /// of `ClassBuilder` and must return it.
     pub modifier: Option<String>,
-    pub flags: Option<String>
+    pub flags: Option<String>,
 }
 
 #[derive(Debug)]
Diff in /home/runner/work/ext-php-rs/ext-php-rs/crates/macros/src/class.rs at line 119:
     let ItemStruct { ident, .. } = &input;
     let class_name = args.name.unwrap_or_else(|| ident.to_string());
     let struct_path = ident.to_string();
-    let flags = args.flags.map(|flags|flags.to_token_stream().to_string());
+    let flags = args.flags.map(|flags| flags.to_token_stream().to_string());
     let class = Class {
         class_name,
         struct_path,
Error: The process '/home/runner/.cargo/bin/cargo' failed with exit code 1
```

Could you fix them, please?
It might also be advisable to rebase your changes on top of the current `master` branch to make the checks pass.

PR #156, opened 2022-09-21
Specify classes as fully-qualified names in stubs

joehoyle:
When stubs are generated, the type annotations don't use a loading `\`, which means they are interpreted as relative to the current namespace. That's wrong, as all types are relative to the root namespace.

### DIFF
--- a/src/describe/stub.rs
+++ b/src/describe/stub.rs
@@ -153,6 +153,7 @@ impl ToStub for Parameter {
 
 impl ToStub for DataType {
     fn fmt_stub(&self, buf: &mut String) -> FmtResult {
+        let mut fqdn = "\\".to_owned();
         write!(
             buf,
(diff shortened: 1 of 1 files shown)


striezel:
The `rustfmt` check is complaining about those changes:
```
/home/runner/.cargo/bin/cargo fmt --all -- --check
Diff in /home/runner/work/ext-php-rs/ext-php-rs/src/describe/stub.rs at line 166:
                 DataType::Object(Some(ty)) => {
                     fqdn.push_str(ty);
                     fqdn.as_str()
-                },
+                }
                 DataType::Object(None) => "object",
                 DataType::Resource => "resource",
                 DataType::Reference => "reference",
Error: The process '/home/runner/.cargo/bin/cargo' failed with exit code 1
```

PR #158, opened 2022-09-28
attempt to fix errors related to `clap`

striezel:
This should fix the errors from the build at
<https://github.com/davidcole1340/ext-php-rs/actions/runs/3145521955/jobs/5112909446>:

       Compiling clap v4.0.0
    error: Unknown `#[clap(long)]` attribute (`#[arg(long)] exists)
      --> crates/cli/src/lib.rs:92:12
       |
    92 |     #[clap(long)]
       |            ^^^^

    error: Unknown `#[clap(long)]` attribute (`#[arg(long)] exists)
       --> crates/cli/src/lib.rs:115:12
        |
    115 |     #[clap(long)]
        |            ^^^^

    error: Unknown `#[clap(short)]` attribute (`#[arg(short)] exists)
       --> crates/cli/src/lib.rs:134:12
        |
    134 |     #[clap(short, long)]
        |            ^^^^^

    error: could not compile `cargo-php` due to 3 previous errors
    warning: build failed, waiting for other jobs to finish...
    Error: Process completed with exit code 101.

These errors were probably caused by the release of [`clap` 4.0.0](https://crates.io/crates/clap/4.0.0) which happened a few hours ago today.

PR #151, opened 2022-09-10
update actions/checkout in GitHub Actions workflows to v3

striezel:
Updates the `actions/checkout` action used in the GitHub Actions workflows to its newest version.

Changes in [actions/checkout](https://github.com/actions/checkout):

> ## v3.0.2
> - Add input `set-safe-directory`
>
> ## v3.0.1
> - Fixed an issue where checkout failed to run in container jobs due to the new git setting `safe.directory`
> - Bumped various npm package versions
>
> ## v3.0.0
>
> - Update to node 16

As far as I can tell this should all be backwards compatible, so I do not expect any breakage.

PR #159, opened 2022-09-28
ci: run clippy only on stable Rust channel

striezel:
~**Note:** This is _still a draft_, because it should only be merged after #158 was approved and merged.~

---

Sparked by the comments about nightly clippy breaking the build too often (see <https://github.com/davidcole1340/ext-php-rs/pull/151#issuecomment-1261151479>), this change will limit clippy to builds with a stable Rust version only.


striezel:
@TobiasBengtsson: Rebased and ready for review.

PR #161, opened 2022-10-12
Update changelog for latest versions

striezel:
Updates the changelog so that the changes for the newer releases are included, too.


davidcole1340 (approved):
Thanks!

Issue #163, opened 2022-10-15
implementing Interfaces


nikeee:
The docs state:
> `#[implements(ce)]` - Implements the given interface on the class. Can be used multiple times. ce must be a valid Rust expression when it is called inside the `#[php_module]` function.

How do I use this to implement an interface provided by the standard library of PHP? For example, I want to implement [`ArrayAccess`](https://www.php.net/manual/en/class.arrayaccess.php) on my own class.

My current approach is this:
```rs
#[php_class]
#[implements(ext_php_rs::zend::class::ClassEntry::try_find("ArrayAccess").unwrap())]
pub struct Test {
}
```
But the `class` type seems to be private:
```
error[E0603]: module `class` is private
   --> src/lib.rs:205:1
```

If I get this to work, how do I actually implement the methods? Does having the appropriate method names suffice (putting renaming of methods aside)?

If this has a solution, I'd love to contribute this to the docs. I really like this library!

davidcole1340:
I don't think I have actually tried to implement an interface, so I will give it a go and report back. To answer your question regarding `ClassEntry`, it is re-exported via `ext_php_rs::zend::ClassEntry::try_find` (https://docs.rs/ext-php-rs/0.7.4/ext_php_rs/zend/type.ClassEntry.html).

nikeee:
I tried adding am example in #167. Do the docs fail due to the version with the interface class entries not being released?

PR #164, opened 2022-10-15
Add some standard zend interfaces

nikeee:
The zend API also has some standard interfaces it exposes:
https://github.com/php/php-src/blob/c8c09b4aaee7ac447828564c6267e62eb304135b/Zend/zend_interfaces.h#L27-L33
```
extern ZEND_API zend_class_entry *zend_ce_traversable;
extern ZEND_API zend_class_entry *zend_ce_aggregate;
extern ZEND_API zend_class_entry *zend_ce_iterator;
extern ZEND_API zend_class_entry *zend_ce_arrayaccess;
extern ZEND_API zend_class_entry *zend_ce_serializable;
extern ZEND_API zend_class_entry *zend_ce_countable;
extern ZEND_API zend_class_entry *zend_ce_stringable;
```

This surfaced in #163 and should make it possible to implement these interfaces.

Note: Im relatively new to rust.. If I've got something wrong, just drop a line and I'd be happy to address the issue. :)


davidcole1340 (approved):
thank you! when updating `allowed_bindings.rs` the `docsrs_bindings.rs` file needs to be re-generated, but I need to document this somewhere. have done it for you.

PR #167, opened 2022-10-17
Add example that shows how to implement an interface

### LINKED ISSUE #163: implementing Interfaces

nikeee:
I've managed to implement an interface as described in #163 and want to provide an example on how to do it. This example is based on the work in #164.

Closes #163

### DIFF
--- a/guide/src/macros/classes.md
+++ b/guide/src/macros/classes.md
@@ -88,10 +88,11 @@ The following example implements [`ArrayAccess`](https://www.php.net/manual/en/c
 # #![cfg_attr(windows, feature(abi_vectorcall))]
 # extern crate ext_php_rs;
 use ext_php_rs::prelude::*;
-use ext_php_rs::{exception::PhpException, zend::ce, types::Zval};
+use ext_php_rs::{exception::PhpResult, types::Zval, zend::ce};
 
 #[php_class]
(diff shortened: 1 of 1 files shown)


davidcole1340:
Not sure why this is failing. Will try have a look at it over the weekend.

davidcole1340:
This should test fine with the changes in #169. Could you rebase on master?

davidcole1340:
Thanks!

PR #174, opened 2022-11-10
chore(macro): Deprecate global state in macro system

### LINKED ISSUE #131: Implicit `#[php_module]` can be akward
Something I've been wanting to do for a while is to remove support for the implicit `#[php_module]`. As it stands, you currently can just drop this at the bottom of your file and have everything register automatically:

```rust
#[php_module]
pub fn module(module: ModuleBuilder) -> ModuleBuilder {
    module
}
```

Two problems arise quickly with larger projects:

### Order matter

Since macros run in order, you must make sure that the module function is the *last* thing the build sees. If anything is created after the fact, the build will fail with very usefulness error messages. 

### Code splitting / rust modules are harder to use.

In one of my larger projects, I split code into a few files. I have to make sure my main lib.rs imports everything in order for it to build. This in turn interacts makes my main module harder to read when I have conditional features split over multiple files.

## Alternative

Follow the similar pattern which pyo3 does: explicit registration. I think its more predictable:

```rust
#[php_module]
pub fn module(m: &Module) -> Result<()> {
    m.add_class::<MyClass>()?;
    Ok(())
}
```

That will allow `module` to appear anywhere in the project.

davidcole1340:
closes #131


azjezz:
@davidcole1340 what is the status here? anything i can do to help? :)

ptondereau:
> @davidcole1340 what is the status here? anything i can do to help? :)

Testing testing and testing haha. From the author, this is mainly finished

azjezz:
I have actually tested it a bit today :D seems good to me

guide/src/macros/impl.md
```diff
@@ -18,9 +19,8 @@ Class methods can take a `&self` or `&mut self` parameter. They cannot take a
 consuming `self` parameter. Static methods can omit this `self` parameter.
 
 To access the underlying Zend object, you can take a reference to a
-`ZendClassObject<T>` in place of the self parameter, where the parameter is
-annotated with the `#[this]` attribute. This can also be used to return a
-reference to `$this`.
+`ZendClassObject<T>` in place of the self parameter, where the parameter must
+be named `self_`. This can also be used to return a reference to `$this`.
```
ju1ius, line 23: Why not requiring `this` as a name?
Makes it both idiomatic and obvious to me.

NEW_MACROS.md
```diff
@@ -0,0 +1,137 @@
...
+fn some_rust_func() {
+    let x = unsafe { phpinfo() };
+    println!("phpinfo: {x}");
+}
+```
+
+## Startup Function
+
+The `#[php_startup]` macro has been deprecated. Instead, define a function with
+the signature `fn(ty: i32, mod_num: i32) -> i32` and provide the function name
```
ju1ius, line 121: Maybe explain what those numbers mean?

src/convert.rs
```diff
@@ -119,6 +119,9 @@ pub trait IntoZval: Sized {
     /// The corresponding type of the implemented value in PHP.
     const TYPE: DataType;
 
+    /// Whether converting into a [`Zval`] may result in null.
+    const NULLABLE: bool;
```
ju1ius, line 123: What's the rationale behind these `NULLABLE`s in conversion traits?

ju1ius:
I did a bit of basic tests and so far your PR solves one of the big pain points I have with the current macro system:
When you have a modestly large project with some submodules, you can not register functions/classes without having to star-import your subodule into the main lib file.

With this PR, now it is possible to do this:
```rs
// lib.rs
mod sub;

#[php_module]
    sub::register(module)
}
```
```rs
// sub/mod.rs

#[php_function]
fn hello_sub(name: &str) -> String {
    format!("Hello sub, {}!", name)
}

pub fn register(module: ModuleBuilder) -> ModuleBuilder {
    module.function(wrap_function!(hello_sub))
}
```
without having to import all the things into the main module, so big 👍🏻 on this !

crates/macros/src/syn_ext.rs
```diff
@@ -136,6 +136,7 @@ impl DropLifetimes for syn::TypePath {
         if let Some(qself) = &mut self.qself {
             qself.ty.drop_lifetimes();
         }
+        self.path.segments.drop_lifetimes();
```
ju1ius, line 139: Tell me if I'm wrong but if this is the part that removes lifetimes from items, wouldn't it be better to throw an error here ?
Because currently if you try to add a lifetime to a registered class, this is the compiler message:
```
error[E0726]: implicit elided lifetime not allowed here
 --> src/lib.rs:4:8
  |
4 | struct Foo<'a> {
  |        ^^^ expected lifetime parameter
  |
  = note: assuming a `'static` lifetime...
help: indicate the anonymous lifetime
  |
4 | struct Foo<'_><'a> {
  |           ++++
```
Which is rather misleading.
Given that the restrictions on lifetime and generic parameters are easily overlooked by newcomers, I think that an error stating explicitely that this is not allowed by the library would be better.

src/builders/module.rs
```diff
@@ -1,11 +1,15 @@
 use crate::{
+    builders::ClassBuilder,
+    class::RegisteredClass,
+    constant::IntoConst,
     error::Result,
     ffi::{ext_php_rs_php_build_id, ZEND_MODULE_API_NO},
+    flags::MethodFlags,
```
ju1ius, line 7: unused import

ju1ius:
This PR fails to build when the `closure` feature is enabled.

```
error[E0046]: not all trait items implemented, missing: `BUILDER_MODIFIER`, `EXTENDS`, `IMPLEMENTS`, `method_builders`, `constructor`, `constants`
   --> /ext-php-rs/src/closure.rs:152:1
    |
152 | impl RegisteredClass for Closure {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `BUILDER_MODIFIER`, `EXTENDS`, `IMPLEMENTS`, `method_builders`, `constructor`, `constants` in implementation
    |
   ::: /ext-php-rs/src/class.rs:28:5
    |
28  |     const BUILDER_MODIFIER: Option<fn(ClassBuilder) -> ClassBuilder>;
    |     ---------------------------------------------------------------- `BUILDER_MODIFIER` from trait
...
31  |     const EXTENDS: Option<fn() -> &'static ClassEntry>;
    |     -------------------------------------------------- `EXTENDS` from trait
...
34  |     const IMPLEMENTS: &'static [fn() -> &'static ClassEntry];
    |     -------------------------------------------------------- `IMPLEMENTS` from trait
...
60  |     fn method_builders() -> Vec<(FunctionBuilder<'static>, MethodFlags)>;
    |     --------------------------------------------------------------------- `method_builders` from trait
...
63  |     fn constructor() -> Option<ConstructorMeta<Self>>;
    |     -------------------------------------------------- `constructor` from trait
...
66  |     fn constants() -> &'static [(&'static str, &'static dyn IntoZvalDyn)];
    |     ---------------------------------------------------------------------- `constants` from trait
```

guide/src/macros/classes.md
```diff
@@ -16,11 +16,10 @@ There are also additional macros that modify the class. These macros **must** be
 placed underneath the `#[php_class]` attribute.
 
 - `#[extends(ce)]` - Sets the parent class of the class. Can only be used once.
-  `ce` must be a valid Rust expression when it is called inside the
-  `#[php_module]` function.
+  `ce` must be a function with the signature `fn() -> &'static ClassEntry`.
```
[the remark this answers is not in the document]
ju1ius: I like this change but it's not clear how you'd retrieve the ClassEntry of one of your registered classes.
There should be an example of how to do that in the documentation, i.e:
```rs
#[php_class]
pub struct Parent;

#[php_class]
#[extends(get_parent_class)]
pub struct Child;

fn get_parent_class() -> &'static ClassEntry {
    // ????????
}
```

NEW_MACROS.md
ptondereau: I've done a small explanation here

ptondereau:
Green CI here! Maybe have a second round of review @ju1ius @joehoyle @joelwurtz @danog

joelwurtz:
I was in fact looking into doing that in a different way as the current and this macro system does not support having a module load in embed. Let me explain : 

PHP rely on a `zend_module_entry` struct for it's module. When writing an extension in C, it's usually done by writing a static object into the root and add an extern on it.

There is also a `get_module` function that returns a pointer to this struct, which allows to know how to load a module when using dynamic loading

In ext-php-rs, only the `get_module` function is declared which means that the extension can only be load if it's used as a dynamic module (you can see there is a hacks in current test method which calls this function when launching an Embed, but unfortunately this hacks does not work for classes as PHP already have loaded thoses classes)

My goal was to write a macro which would mimic the way of PHP, something like : 

```rust
php_module! {
    startup_function = my_startup_fonction,
    name = "extension-name",
    ...
};
```

Which would done the following : 

```
pub extern "C" const module_extension: zend_module_entry = zend_module_entry {
    ...
}

#[cfg(lib)]
pub extern "C" fn get_module() -> *const zend_module_entry {
    return &zend_module_entry;
}
```

This way the get_module will only be available when compiled as a lib but not as a binary where it would be loaded because there is an extern on the zend_module_entry

ju1ius:
@joelwurtz I'm not sure I understand...

I have the feeling that you're confusing `modules` with `extensions`.

Modules are loaded from ini entries (`extension=libfoo.so`) and must export a `get_module` symbol which must be function returning a `zend_module_entry` pointer.

Extensions are loaded from ini entries (`zend_extension=libbar.so`) and must export a `zend_extension_entry` symbol which must be a statically allocated `zend_extension` struct.

So to me it sounds like what you'd like to implement is the hybrid extension pattern [described in the book](https://www.phpinternalsbook.com/php7/extensions_design/zend_extensions.html#hybrid-extensions).

Or maybe I'm just missing something, in which case would you mind clarifying? 😉

joelwurtz:
I was correctly talking about a PHP extension (zend_module_entry) but i misunderstood how sapi (and so embed) loaded themself as a PHP extension (so you could build a php binary with extra functions)

But in fact this is done when using the `php_module_startup` function (https://github.com/php/php-src/blob/master/main/main.c#L2027) which takes an sapi module and add optional `zend_module_entry` pointer (so it can register a new extension when starting up)

As an example this is done by php-fpm to register it's own functions (like it's shutdown one) https://github.com/php/php-src/blob/master/sapi/fpm/fpm/fpm_main.c#L774

So disregard my comment it's off topic for this PR, sorry for the noise

I was confused on how PHP builtin extension (like date or other ones) were loaded (when they are not a shared file), and in fact it's just a template header file that is computed during make which list all extensions that will be compiled directly inside binaries (php cli / fpm / embed / etc ...) see https://github.com/php/php-src/blob/master/main/internal_functions.c.in

NEW_MACROS.md
```diff
@@ -0,0 +1,140 @@
...
+`mod_num` is the module number assigned by PHP for its internal management. The `ModuleBuilder` always defaults this parameter to `0`.
+`ty` is the type of module (flag) which can be set to `MODULE_PERSISTENT = 0`.
+The return number is for internal purposes and should always be `0` in this case. It's an internal mechanism to tell to the ModuleBuilder weither to use the defautlt startup function or the overriden one. (see: `crates/macros/src/module.rs:L35`)
+
+```rs
+use ext_php_rs::prelude::*;
+
+fn startup(ty: i32, mod_num: i32) -> i32 {
+    // register extra classes, constants etc
+    5.register_constant("SOME_CONST", mod_num);
```
joelwurtz, line 132: typo ? i believe it should be in the get_module function

joelwurtz (commented):
About this move i'm :+1: even if it's add boilerplate i find this just as verbose as it should to be self understandable with code without too much work

NEW_MACROS.md
ptondereau: I think not. It's the global constant declaration here, and it's mostly used as a flag for the PHP end user
see: 
https://github.com/davidcole1340/ext-php-rs/blob/417f495090178cc07df9d00d827d0ee57fb59bef/src/constant.rs#L15-L23

ju1ius: > typo ? i believe it should be in the get_module function

No this code is correct.

The `get_module` function is used for registering functions, ini settings and module metadata (name, version, dependencies). Everything else must happen in `startup` (AKA `MINIT`), unless it need to be allocated per-request, in which case one must use `request_startup` (AKA `RINIT`).

guide/src/macros/classes.md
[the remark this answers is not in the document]
ju1ius: Answering to my past self: `<Child as RegisteredClass>::get_metadata().ce()`

guide/src/macros/impl.md
ju1ius: My future self is still sad about using `self_` instead of the idiomatic `this`...

NEW_MACROS.md
ju1ius: Off the top of my head the module type can only be one of `MODULE_PERSISTENT` or `MODULE_TEMPORARY`, so it would be nicer to have a :
```rs
pub enum ModuleType {
    Persistent,
    Temporary,
}
impl From<i32> for ModuleType {
    fn from(value: i32) -> Self {
        match value => {
            ffi::MODULE_PERSISTENT => Persistent,
            ffi::MODULE_TEMPORARY => Temporary,
            _ => panic!("onoes!")
        }
    }
}
// TODO: impl From<ModuleType> for i32
```

and then:

```rs
fn startup(ty: ModuleType, id: i32) -> Result<(), E> {
    Ok(())
}
```

crates/macros/src/syn_ext.rs
ju1ius: Answering my past self: this part drops lifetimes of function and method arguments.

Since I don't agree with the type conversion magic at function boundaries, I have no opinion on whether dropping lifetime for arguments is a good thing or not.

But I still think that `#[php_class]` should fail to compile with an error message like
````
Lifetimes and generic parameters are not allowed on classes and class members
```

instead of the message seen in the [previous comment](https://github.com/davidcole1340/ext-php-rs/pull/174/files#diff-d69bdb20c8c49926e0b3f843284710b84915fb209e97d120290136743587607d).

NEW_MACROS.md
ptondereau: Temporary remains a rare case, and then the default case is totally fine for now.
Let's open another PR, after merging it

ju1ius:
> I was correctly talking about a PHP extension

Yes the fact that these terms are mixed and matched in the source PHP source code and documentation is rather unfortunate (mostly due to legacy I guess) and doesn't ease communication. I personnaly think that «modules vs extensions» is clearer than «PHP extensions vs Zend extensions», but to each his own I guess.

## WRT the actual PR

Sorry, I don't have the bandwidth to make an exhaustive review but from what I can remember from last year it was a clear step forward to me, so I'm all in favour of merging it while it's hot !

NEW_MACROS.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+## Startup Function
+
+The `#[php_startup]` macro has been deprecated. Instead, define a function with
+the signature `fn(ty: i32, mod_num: i32) -> i32` and provide the function name
+to the `#[php_module]` attribute:
+`mod_num` is the module number assigned by PHP for its internal management. The `ModuleBuilder` always defaults this parameter to `0`.
+`ty` is the type of module (flag) which can be set to `MODULE_PERSISTENT = 0`.
+The return number is for internal purposes and should always be `0` in this case. It's an internal mechanism to tell to the ModuleBuilder weither to use the defautlt startup function or the overriden one. (see: `crates/macros/src/module.rs:L35`)
```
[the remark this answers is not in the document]
ptondereau: It could be done in seperate PR, it will require some parsing here to enforce this signature

Xenira:
Closing this as superseded by #372

PR #195, opened 2022-11-24
fixes CI workflow configuration

ju1ius:
The latest runs were broken for some reason so I fixed it.
Here's a summary of what this PR does:

* upgrades LLVM to v14
* migrates from the unmaintained `action-rs/*` actions to [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain), using [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache/) as a cache layer.
* adds a cache layer for LLVM build
* adds a weekly cron schedule for all workflows
* fixes an issue in the docblocks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,8 +12,9 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        php: ['8.0', '8.1']
+        php: ["8.0", "8.1"]
         rust: [stable, nightly]
+        clang: ["14"]
         phpts: [ts, nts]
         exclude:
           # ext-php-rs requires nightly Rust when on Windows.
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -34,59 +37,67 @@ jobs:
...
+        # Uncomment the following if statement if caching nightly deps
+        # ends up causing too much cache invalidation.
+        # if: matrix.rust == 'stable'
+        with:
+          # increment this manually to force cache eviction
+          prefix-key: "v0-rust"
+      # LLVM & Clang
+      - name: Cache LLVM and Clang
+        id: cache-llvm
+        uses: actions/cache@v2
```
ptondereau, line 56: cache@v3 isn't available?

ptondereau (approved):

ptondereau:
One thing I like to do is to add cron every week for building + doc. That way, we can anticipate that kind of error

.github/workflows/build.yml
ju1ius: Good catch !
I will change that.

ptondereau:
Thank you!

PR #196, opened 2022-11-24
Add get_id() and hash() methods on ZendObject

ju1ius:
The `get_id` method is strictly equivalent to calling `spl_object_id()` with the object as parameter.

The `get_hash` method is strictly equivalent to calling `spl_object_hash()` with the object as parameter.

### DIFF
--- a/src/types/object.rs
+++ b/src/types/object.rs
@@ -196,6 +196,29 @@ impl ZendObject {
         T::from_zend_object(self)
     }
 
+    /// Returns an unique identifier for the object.
+    ///
+    /// The id is guaranteed to be unique for the lifetime of the object.
+    /// Once the object is destroyed, it may be reused for other objects.
+    /// This is equivalent to calling the [`spl_object_id`] PHP function.
+    ///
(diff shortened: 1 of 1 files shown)


src/types/object.rs
```diff
@@ -196,6 +196,29 @@ impl ZendObject {
...
+    }
+
+    /// Returns an unique hash for the object.
+    ///
+    /// The hash is guaranteed to be unique for the lifetime of the object.
+    /// Once the object is destroyed, it may be reused for other objects.
+    /// This is equivalent to calling the [`spl_object_hash`] PHP function.
+    ///
+    /// [`spl_object_hash`]: https://www.php.net/manual/function.spl-object-hash.php
+    pub fn get_hash(&self) -> String {
```
ju1ius, line 218: Not sure about the naming here. I think it should be `hash()` not `get_hash()` since it returns a new string on each call. WDYT?

ptondereau: I would go for `hash()` to be consistent with the root method

ptondereau:
@ju1ius you can rebase

PR #201, opened 2022-11-26
chore: use php-discovery to find matching PHP build

azjezz:
I currently have over 18 PHP installations locally, having to swipe between them for different projects is kinda annoying, so here's my solution for the problem :)

[`php-discovery`](https://github.com/azjezz/php-discovery) will attempt to find all PHP builds in the system, even if not listed in `PATH`, and create a struct representing each build, containing all information necessary for ext-php-rs.

note: ext-php-rs will still have to do the work to get `includes` on windows, while i can certainly add it as a feature in [`php-discovery`](https://github.com/azjezz/php-discovery), i feel that it doesn't belong, and that it should do only 1 job, which is discover available PHP builds.

### DIFF
--- a/windows_build.rs
+++ b/windows_build.rs
@@ -7,7 +7,6 @@ use std::{
 };
 
 use anyhow::{Context, Result};
-use php_discovery::build::Architecture;
 use php_discovery::build::Build;
 
 use crate::PHPProvider;
--- a/windows_build.rs
+++ b/windows_build.rs
@@ -1,20 +1,21 @@
+use std::io::Write;
 use std::{
-    convert::TryFrom,
-    fmt::Display,
-    io::{Cursor, Read, Write},
+    io::{Cursor, Read},
     path::{Path, PathBuf},
     process::Command,
     sync::Arc,
 };
 
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result};
+use php_discovery::build::Architecture;
+use php_discovery::build::Build;
 
-use crate::{PHPInfo, PHPProvider};
+use crate::PHPProvider;
 
 const USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
 
 pub struct Provider<'a> {
-    info: &'a PHPInfo,
+    build: &'a Build,
     devel: DevelPack,
 }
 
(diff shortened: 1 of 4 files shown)


windows_build.rs
```diff
@@ -32,18 +34,22 @@ impl<'a> Provider<'a> {
...
-        let version = info.version()?;
-        let is_zts = info.thread_safety()?;
-        let arch = info.architecture()?;
-        let devel = DevelPack::new(version, is_zts, arch)?;
+    fn new(build: &'a Build) -> Result<Self> {
+        // don't use `build.version.to_string()` as it includes extra part which is not needed.
+        let version = format!(
+            "{}.{}.{}".build.version.major,
+            build.version.minor, build.version.release
+        );
```
azjezz, line 42: `version.to_string()` could return something like `8.2.0RC6`, which is not something we want.

ptondereau:
@azjezz thank you! This indeed remove a lot of code from build. I've approved the CI workflow

azjezz:
ah! forgot to define `php81`!

windows_build.rs
```diff
@@ -32,18 +34,23 @@ impl<'a> Provider<'a> {
...
-    fn new(info: &'a PHPInfo) -> Result<Self> {
-        let version = info.version()?;
-        let is_zts = info.thread_safety()?;
-        let arch = info.architecture()?;
-        let devel = DevelPack::new(version, is_zts, arch)?;
+    fn new(build: &'a Build) -> Result<Self> {
+        // don't use `build.version.to_string()` as it includes extra part which is not
+        // needed.
+        let version = format!(
+            "{}.{}.{}".build.version.major,
```
ptondereau, line 41:
```suggestion
            "{}.{}.{}", build.version.major,
```

ju1ius:
Hi @azjezz,

While I appreciate your motivation and considerable efforts to improve the current situation, having a script taking ~10sec to [scan my entire `/usr` directory](https://github.com/azjezz/php-discovery/blob/0627c99157523ae2636eebc14e3a3f2c89b27294/src/lib.rs#L69), only to find exactly the same thing that `which php` finds in 5ms - that is to say not the right version because I keep all my custom PHP builds under my home directory - is not something I think I'll enjoy very much.

I'd rather have a simple environment variable override like `PHP_PREFIX=<prefix>` where `<prefix>` is the `--prefix` option passed to `./configure` (which would imply `<prefix>/bin/php` and `<prefix>/bin/php-config` on unixes).
If `PHP_PREFIX` is not set, then just fallback to `which php-config`.

But hey, that's only my personal taste. 😉

ptondereau:
> Hi @azjezz,
> 
> 
> 
> While I appreciate your motivation and considerable efforts to improve the current situation, having a script taking ~10sec to [scan my entire `/usr` directory](https://github.com/azjezz/php-discovery/blob/0627c99157523ae2636eebc14e3a3f2c89b27294/src/lib.rs#L69), only to find exactly the same thing that `which php` finds in 5ms - that is to say not the right version because I keep all my custom PHP builds under my home directory - is not something I think I'll enjoy very much.
> 
> 
> 
> I'd rather have a simple environment variable override like `PHP_PREFIX=<prefix>` where `<prefix>` is the `--prefix` option passed to `./configure` (which would imply `<prefix>/bin/php` and `<prefix>/bin/php-config` on unixes).
> 
> If `PHP_PREFIX` is not set, then just fallback to `which php-config`.
> 
> 
> 
> But hey, that's only my personal taste. 😉 
> 
> 

Yes sometimes, I need to test with different compiled version of php (debug) and I know where to find it. In a build context, an environment variable could be more efficient.

ju1ius:
> having a script taking ~10sec to scan my entire /usr directory,

I stand corrected, it does perform some globbing so it's more like under a second, which is much better than I tought initially.

Still won't do anything useful in my particular case, but simplifying the build workflow is always nice ! 👍🏻

azjezz:
i have added a soft-criteria for PHP version, which can be used to select which binary to use, for example, here i can select `8.1.0-dev` ( i say `soft` because it won't fail if no matching build is found, but rather continue to grab the first one with a matching PHP API )
image

azjezz:
( sorry about the repeated failure on windows, not testing it ^^" )

azjezz:
passed ✔️

azjezz:
^ removed unused import, won't have any effect on the build result

ptondereau (approved):
LGTM! Thank you!

PR #202, opened 2022-11-26
feat: allows ZendStr to contain null bytes

### LINKED ISSUE #200: ZendStr should accept strings containing null bytes.
Hi,

PHP strings are allowed to contain null bytes, but the current implementation of ZendStr doesn't allow it.

```rs
#[php_function]
pub fn class_name_of(obj: &ZendObject) -> String {
    obj.get_class_name().unwrap()
}
```
```php
<?php

var_dump(class_name_of(new class {}));
```
Expected result: 
```
string(%d) "class@anonymous/path/to/file.php:3$0"
// or, with XDebug enabled:
string(%d) "class@anonymous\000/path/to/file.php:3$0"
```
Actual result:
```
PHP Fatal error: Uncaught Exception: String given contains NUL-bytes which cannot be present in a C string. [...]
```

ju1ius:
Closes #200

## Rationale
In PHP zend_strings are binary strings with no encoding information. They can contain any byte at any position.
The current implementation use `CString` to transfer zend_strings between Rust and PHP, which prevents zend_strings containing null-bytes to roundtrip through the ffi layer. Moreover, `ZendStr::new()` accepts only a `&str`, which is incorrect since a zend_string is not required to be valid UTF8.

When reading the PHP source code, it is apparent that  most functions marked with `ZEND_API` that accept a `const *char` are convenience wrappers that convert the `const *char` to a zend_string and delegate to another function. For example [zend_throw_exception()](https://github.com/php/php-src/blob/eb83e0206c9b9261d786943bf2c5ad61dca287e2/Zend/zend_exceptions.c#L823) takes a `const *char message`, and just converts it to a zend_string before delegating to [zend_throw_exception_zstr()](https://github.com/php/php-src/blob/eb83e0206c9b9261d786943bf2c5ad61dca287e2/Zend/zend_exceptions.c#L795).

I kept this PR focused around `ZendStr` and it's usages in the library, but it should be seen as the first step of a more global effort to remove usages of `CString` everywhere possible.

Also, I didn't change the return type of the string related methods of `Zval` (e.g. I could have made `Zval::set_string()` 
 accept an `impl AsRef<[u8]>` instead of `&str` and return `()` instead of `Result<()>`). If I get feedback that it should be done in this PR, I'll do it.

### ZendStr
* [BC break]: `ZendStr::new()` and `ZendStr::new_interned()` now accept an `impl AsRef<[u8]>` instead of just `&str`, and are therefore infaillible (outside of the cases where we panic, e.g. when allocation fails). This is a BC break, but it's impact shouldn't be huge (users will most likely just have to remove a bunch of `?` or add a few `Ok()`).
* [BC break]: Conversely, `ZendStr::as_c_str()` now returns a `Result<&CStr>` since it can fail on strings containing null bytes.
* [BC break]: `ZensStr::as_str()` now returns a `Result<&str>` instead of an `Option<&str>` since we have to return an error in case of invalid UTF8.
* adds method `ZendStr::as_bytes()` to return the underlying byte slice.
* adds convenience methods `ZendStr::as_ptr()` and `ZendStr::as_mut_ptr()` to return raw pointers to the zend_string.

 ### ZendStr conversion traits
* adds `impl AsRef<[u8]> for ZendStr`
* [BC break]: replaces `impl TryFrom<String> for ZBox<ZendStr>` by `impl From<String> for ZBox<ZendStr>`.
* [BC break]: replaces `impl TryFrom<&str> for ZBox<ZendStr>` by `impl From<&str> for ZBox<ZendStr>`.
* [BC break]: replaces `impl From<&ZendStr> for &CStr` by `impl TryFrom<&ZendStr> for &CStr`.

### Error
* adds new enum member `Error::InvalidUtf8` used when converting a `ZendStr` to `String` or `&str`

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -16,6 +16,7 @@ use crate::{
     convert::{FromZval, IntoZval},
     error::{Error, Result},
     ffi::{
+        ext_php_rs_is_known_valid_utf8, ext_php_rs_set_known_valid_utf8,
         ext_php_rs_zend_string_init, ext_php_rs_zend_string_release, zend_string,
         zend_string_init_interned,
     },
@@ -293,11 +294,13 @@ impl ZendStr {
     /// assert!(s.as_str().is_ok());
     /// ```
     pub fn as_str(&self) -> Result<&str> {
-        // TODO: we should avoid revalidating the string repeatedly:
-        // check for `GC_FLAGS(self) & IS_STR_VALID_UTF8` and then:
-        // * true => return std::str::from_utf8_unchecked(...)
-        // * false => return std::str::from_utf8(...) and set the flag on success
-        std::str::from_utf8(self.as_bytes()).map_err(|_| Error::InvalidUtf8)
+        if unsafe { ext_php_rs_is_known_valid_utf8(self.as_ptr()) } {
+            let str = unsafe { std::str::from_utf8_unchecked(self.as_bytes()) };
+            return Ok(str);
+        }
+        let str = std::str::from_utf8(self.as_bytes()).map_err(|_| Error::InvalidUtf8)?;
+        unsafe { ext_php_rs_set_known_valid_utf8(self.as_ptr() as *mut _) };
+        Ok(str)
     }
 
     /// Returns a reference to the underlying bytes inside the Zend string.
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -77,6 +77,8 @@ impl ZendStr {
     /// ```
     pub fn new(str: impl AsRef<[u8]>, persistent: bool) -> ZBox<Self> {
         let s = str.as_ref();
+        // TODO: we should handle the special cases when length is either 0 or 1
+        // see `zend_string_init_fast()` in `zend_string.h`
         unsafe {
             let ptr = ext_php_rs_zend_string_init(s.as_ptr().cast(), s.len(), persistent)
                 .as_mut()
@@ -286,6 +288,10 @@ impl ZendStr {
     /// assert!(s.as_str().is_ok());
     /// ```
     pub fn as_str(&self) -> Result<&str> {
+        // TODO: we should avoid revalidating the string repeatedly:
+        // check for `GC_FLAGS(self) & IS_STR_VALID_UTF8` and then:
+        // * true => return std::str::from_utf8_unchecked(...)
+        // * false => return std::str::from_utf8(...) and set the flag on success
         std::str::from_utf8(self.as_bytes()).map_err(|_| Error::InvalidUtf8)
     }
 
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -30,7 +30,7 @@ use crate::{
 /// cannot represent unsized types, an array of size 1 is used at the end of the
 /// type to represent the contents of the string, therefore this type is
 /// actually unsized. All constructors return [`ZBox<ZendStr>`], the owned
-/// varaint.
+/// variant.
 ///
 /// Once the `ptr_metadata` feature lands in stable rust, this type can
 /// potentially be changed to a DST using slices and metadata. See the tracking issue here: <https://github.com/rust-lang/rust/issues/81513>
@@ -46,19 +46,13 @@ static INTERNED_LOCK: Mutex<()> = const_mutex(());
 // on the alias `ZendStr` :( <https://github.com/rust-lang/rust-clippy/issues/7702>
 #[allow(clippy::len_without_is_empty)]
 impl ZendStr {
-    /// Creates a new Zend string from a [`str`].
+    /// Creates a new Zend string from a slice of bytes.
     ///
     /// # Parameters
     ///
     /// * `str` - String content.
     /// * `persistent` - Whether the string should persist through the request
     ///   boundary.
-    ///
-    /// # Returns
-    ///
-    /// Returns a result containing the Zend string if successful. Returns an
-    /// error if the given string contains NUL bytes, which cannot be
-    /// contained inside a C string.
     ///
     /// # Panics
     ///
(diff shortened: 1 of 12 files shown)


ju1ius:
A few off-topic remarks:
1. since persistent zend_strings are almost never what you want, and there is a `ZendStr::new_interned` constructor, shouldn't  we add a `ZendStr::new_persistent()` constructor and remove the `persistent` parameter in  `ZendStr::new()` ?
2. we don't currently tackle the special cases: when the length of the string is either 0 or 1 like in [zend_string_init_fast()](https://github.com/php/php-src/blob/eb83e0206c9b9261d786943bf2c5ad61dca287e2/Zend/zend_string.h#L179)
3. I've seen a few well-known frameworks reserving a memory buffer for their error handlers ([like so](https://github.com/symfony/symfony/blob/c79d4ab7e787eb27d83581ccb8e15a4e11e88df3/src/Symfony/Component/ErrorHandler/ErrorHandler.php#L110)). I guess the goal is for them to be able to run the handlers even when the allowed memory limit is reached. I don't know if this applies for extensions, but maybe panicking on failed allocations is not good citizenship?

WDYT?

ptondereau (approved):

ju1ius:
@ptondereau  I overlooked some of the `impls`, so this still need a bit of work.

ptondereau:
> 1. since persistent zend_strings are almost never what you want, and there is a ZendStr::new_interned constructor, shouldn't we add a ZendStr::new_persistent() constructor and remove the persistent parameter in ZendStr::new() ?

IMHO, this is a code smell to bring a builder, but I agree that we can have a specific constructor for a not so used context.

> 2. we don't currently tackle the special cases: when the length of the string is either 0 or 1 like in [zend_string_init_fast()](https://github.com/php/php-src/blob/eb83e0206c9b9261d786943bf2c5ad61dca287e2/Zend/zend_string.h#L179)

It should be done then. 

> 3. I've seen a few well-known frameworks reserving a memory buffer for their error handlers ([like so](https://github.com/symfony/symfony/blob/c79d4ab7e787eb27d83581ccb8e15a4e11e88df3/src/Symfony/Component/ErrorHandler/ErrorHandler.php#L110)). I guess the goal is for them to be able to run the handlers even when the allowed memory limit is reached. I don't know if this applies for extensions, but maybe panicking on failed allocations is not good citizenship?

I guess PHP will crash before Rust have the allocation state and this leads to an unresolved error (that I need to create issue) that leak memory of Rust's allocated stuff. I would go for a wrapping with [`zend_try+zend_catch` ](https://flylib.com/books/en/2.565.1/dealing_with_errors.html)in our C wrapper, but maybe this isn't the best solution.

src/types/string.rs
```diff
@@ -266,29 +268,30 @@ impl ZendStr {
...
-    /// let s = ZendStr::new("hello, world!", false).unwrap();
-    /// let as_str = s.as_str();
-    /// assert_eq!(as_str, Some("hello, world!"));
+    /// let s = ZendStr::new("hello, world!", false);
+    /// assert!(s.as_str().is_ok());
     /// ```
-    pub fn as_str(&self) -> Option<&str> {
-        self.as_c_str().to_str().ok()
+    pub fn as_str(&self) -> Result<&str> {
+        std::str::from_utf8(self.as_bytes()).map_err(|_| Error::InvalidUtf8)
```
ju1ius, line 289: I'm wary that we might have a Schlemiel the Painter problem here.
If I remember correctly, there's a flag somewhere at the C-level to indicate that the string has already been utf8-validated (i.e. by the mbstring or pcre extensions), need to investigate...

ju1ius: It's [here](https://github.com/php/php-src/blob/master/ext/mbstring/mbstring.c#L4369) and also [here](https://github.com/php/php-src/blob/master/ext/pcre/php_pcre.c#L1469).
So we could do something like:
```rs
if GC_FLAGS(self) & IS_STR_VALID_UTF8 {
   return std::str::from_utf8_unchecked(...);
}
let s = std::str::from_utf8(self.as_bytes()).map_err(|_| Error::InvalidUtf8)?;
GC_ADD_FLAG(self, IS_STR_VALID_UTF8):
return s;
```

ju1ius: This requires adding stuff to the ffi bindings though (the gc flags related stuff in `zend_types.h`).
I'm going to leave this off for now and do it in a separate PR.

ju1ius:
> IMHO, this is a code smell to bring a builder, but I agree that we can have a specific constructor for a not so used context.
 
The idea wasn't to bring a builder, just to remove the `persistent` parameter and introduce a new constructor function:

```rs
impl ZendStr {
  pub fn new(value: impl AsRef<[u8]>) -> Self {}
  pub fn new_persistent(value: impl AsRef<[u8]>) -> Self {}
  pub fn new_interned(value: impl AsRef<[u8]>) -> Self {}
}
```
This is more a DX improvement. For example this would allow removing the `persistent` parameter from the conversion traits (`FromZval` et al.). Since it is only ever use for strings and even then rarely so, I find it cumbersome to have to specify it hen it is never needed (like `true.into_zval(false)`...).

But this is not needed for this PR, so let's keep things the way they are for now.

ju1ius:
> > 3\. we don't currently tackle the special cases: when the length of the string is either 0 or 1 like in [zend_string_init_fast()](https://github.com/php/php-src/blob/eb83e0206c9b9261d786943bf2c5ad61dca287e2/Zend/zend_string.h#L179)
> 
> It should be done then.

This requires adding symbols to the ffi bindings (see [related constants in zend_string.h](https://github.com/php/php-src/blob/eb83e0206c9b9261d786943bf2c5ad61dca287e2/Zend/zend_string.h#L58)).
I'm going to leave this off for now and do it in a separate PR.

src/types/string.rs
ptondereau: Dors this require mbstring to be installed

ju1ius: > Dors this require mbstring to be installed

No, the flag is _used_ by `mbstring` and `PCRE` but it is defined in [zend_types.h](https://github.com/php/php-src/blob/eb83e0206c9b9261d786943bf2c5ad61dca287e2/Zend/zend_types.h#L721)

ju1ius: Yes I think so too!
I proposed to do this in another PR to keep this one focused, but if you think it should be done here I'm OK with that.
Do you confirm?

ptondereau: Yes, this was my intention. Sorry and thank you

ju1ius: It shall be done then! 😉

ptondereau (approved):
LGTM!

PR #210, opened 2022-11-30
honour PHP_CONFIG & rebuild automatically when env vars change

### LINKED ISSUE #208: PHP_CONFIG env var not honored by build script
Hi,

`cargo php install` supports reading the location of `php-config` from the `PHP_CONFIG` environment variable.

However, the build script does not honour this variable and always search for `php-config` in the PATH.

This can result in build conflicts when the `php-config` binary does not match the `php` binary.

### LINKED ISSUE #209: cargo build should track environment variables
Hi,

Currently when we want to switch the PHP version for a build, we have to run
```sh
cargo clean
PATH="/path/to/build:${PATH}" cargo build
```
Which wipes out the entire compilation cache.

The build script should use `cargo:rerun-if-env-changed=PATH` so that we can just do:
```sh
PATH="/path/to/build:${PATH}" cargo build
```
and cargo rebuilds only what's needed, not the entire dependency tree.

ju1ius:
Closes #208.
Closes #209.

Please see the linked issues for a rationale.

### Build scripts
* the `unix_build.rs` script now honours the `PHP_CONFIG` environment variable, like `cargo php install`
* use `cargo:rerun-if-env-changed` for the `PHP`, `PHP_CONFIG` and `PATH` environment variables, in order to avoid needless recompilations of the whole dependency tree.

### Documentation
While trying to document the aforementioned changes, I realized that there was no chapter about installing and setting up a PHP environment to develop PHP extensions. So I refactored the first chapters of the book into a `Getting Started` section, including instructions on how to quickly setup a PHP environment.

Do not hesitate to point-out spelling/grammar mistakes as I'm not a native english speaker.


ptondereau (approved):

joehoyle:
I think this might have caused an issue for me. I'm seeing 

```
     Fresh skeptic v0.13.7
   Compiling ext-php-rs v0.10.0
     Running `.../debug/build/ext-php-rs-a3bbbf43415ad934/build-script-build`
```

On every `cargo check` / `cargo build` which takes maybe 3+ seconds, which also makes Rust-Analyzer super slow. Is the intention here that if the env vars don't change, the build should still be cached? I had to switch back to 0.8.3 where I get fully cached builds again

ptondereau:
@ joehoyle hey can you submit a ticket please, I'll take a look! BTW while working on it, I didn't see that regression

ju1ius:
> Is the intention here that if the env vars don't change, the build should still be cached?

Yes, that's how  `cargo:rerun-if-env-changed` is supposed to work, unless something else is interfering...

ju1ius:
Found [a 3 years old issue](https://github.com/rust-lang/cargo/issues/7432) on the cargo repo that _might_ be related.
Maybe it's worth investigating?

Issue #219, opened 2023-01-30
Zval type flags are incorrect for IS_CALLABLE / IS_INDIRECT


joehoyle:
While trying to working with `ExecutorGlobals::get().symbol_table` where the PHP global variables can be `INDIRECT` Zvals. (See https://wiki.php.net/phpng-int for a description of INDIRECT Zvals, specifically around their use in symbol tables.

ext-php-rs is detecting the type of `INDIRECT` zval's as `Callable`, which is not correct. Looking at the PHP src at https://github.com/php/php-src/blob/master/Zend/zend_types.h#L562-L575 there is overlap in `IS_CALLABLE` and `IS_INDIRECT` (similarly `IS_PTR` and `IS_ITERABLE`). For some reason ext-php-rs has chosen to take the flag `IS_CALLABLE` in https://github.com/davidcole1340/ext-php-rs/blob/master/src/flags.rs#L254, when it should actually be checking `IS_INDIRECT` and treating it like a pointer to another Zval.

I don't believe Zval's type flag can actually be `IS_CALLABLE`, as is suggested by the php-src link above, that is a "fake" type.

Switching this out would probably break the current implementation of `Zval::is_callable()` though, so I'm not sure what the correct solution here is exactly.

PR #224, opened 2023-02-05
Add support for Zval IS_INDIRECT

joehoyle:
As mentioned in #219, I believe we're making incorrect use of IS_CALLABLE on Zval types right now. Zval type bits are never actually stored as IS_CALLABLE, which overlaps with the _actual_ type value of IS_INDIRECT. IS_INDIRECT is almost the same as IS_REFERENCE, but is a direct reference to another Zval rather than reference object. As `Zval::is_callable()` and `Zval::callable()` don't actually make use of `IS_CALLABLE` then I think it's safe to switch this out.


ptondereau:
@joehoyle can you please rebase?

joehoyle:
Ok fixed merge conflicts

danog:
Could you please rebase?

joehoyle:
Ok rebazed

PR #225, opened 2023-02-05
Support registering ini definitions for modules

joehoyle:
PHP extensions that want to provide ini settings can do so using `IniEntryDef::register()`; values can then be fetched via `GlobalExecutor::ini_values()`.


ptondereau:
Thank you! Did you tested it?

danog:
Thank you, this is super useful!

PR #227, opened 2023-02-07
Mate GlobalExecutor::get_mut() public

joehoyle:
THis is useful if you want to insert global variables or the like into the the global executor.


ptondereau (approved):
LGTM

PR #232, opened 2023-02-24
Stop watching Cargo.lock for changes

### LINKED ISSUE #220: Crate is rebuilt on every build due to rerun-if-env-changed
As mentioned in https://github.com/davidcole1340/ext-php-rs/pull/210#issuecomment-1406892456, every time I run `cargo check` etc then the ext-php-rs build script is run, due to the use of `rerun-if-env-changed`.

I tracked it down specifically to the `PATH` env var -- `PHP` and `PHP_CONFIG` don't cause issues (though those env vars are not set). My `$PATH` _is_ stable, so it's not like it's changing. (unless the build script it's self is setting something).

rmccue:
When loaded via Cargo, there's no Cargo.lock directly in this crate, so the lack of file existence causes unnecessary rebuilds every time.

Fixes #220.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -237,7 +237,6 @@ fn main() -> Result<()> {
     }
 
     println!("cargo:rerun-if-changed=build.rs");
-    println!("cargo:rerun-if-changed=Cargo.lock");
 
     // docs.rs runners only have PHP 7.4 - use pre-generated bindings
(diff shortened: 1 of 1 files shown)


ptondereau (approved):
Thank you!

PR #240, opened 2023-05-12
Switch to use zend apis for array iteration

### LINKED ISSUE #218: Segment Fault  passing  a vector as argument
I'm having trouble to handle Arrays as parameters, if the array has a string key works fine, but if its a int index a segfault occurs.

```
#[php_class]
pub struct Serial {}

#[php_impl]
impl Serial {
    //pub fn __construct(name: &str, values: Vec<f64>) -> Self {
    //pub fn __construct(name: String, values: Vec<i64>) -> Self {
    //pub fn __construct(name: String, values: Vec<&Zval>) -> Self {
    //pub fn __construct(name: String, values: Vec<String>) -> Self {
    pub fn __construct(name: String, values: &ZendHashTable) -> Self {
        println!("values.len({})", values.len());

        let iter = values.iter().map(|(idx, _key, val)| {
            println!("values.iter() {} -  - {:?}", idx, val);
            val.double().unwrap_or(0.0)
        });

        println!("{:?}", values);

        println!("{:?}", iter.collect::<Vec<f64>>());
    }
}
```

The code bellow does not work
```
$var = new Serial("age", [1,2,3,4]);
//or
$var = new Serial("age", [
        "1" => 1.0,
        "2" => 2.0,
        "3" => 3.0
]);
```
RESULT
```
php -d extension=./target/debug/libphp_teste.dylib test.php
values.len(3)
zsh: segmentation fault  php -d extension=./target/debug/libphp_teste.dylib test.php
```

The code bellow works fine
```
$var = new Serial("age", [
        "a" => 1.0,
        "b" => 2.0,
        "c" => 3.0
]);
```
RESULT
```
values.len(3)
{"a": Zval { type: Double, val: Some(1.0) }, "b": Zval { type: Double, val: Some(2.0) }, "c": Zval { type: Double, val: Some(3.0) }}
values.iter() 9223372036854953478 -  - Zval { type: Double, val: Some(1.0) }
values.iter() 9223372036854953479 -  - Zval { type: Double, val: Some(2.0) }
values.iter() 9223372036854953480 -  - Zval { type: Double, val: Some(3.0) }
[1.0, 2.0, 3.0]
```

## PHP VERSION
```
PHP 8.2.1 (cli) (built: Jan 12 2023 19:14:53) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.1, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.1, Copyright (c), by Zend Technologies
```

joehoyle:
Fixes #218


ptondereau:
Hey @joehoyle thanks for the PR!
By any chance, did you try your code against the example in the associated issue?

PR #249, opened 2023-07-07
Add streams API, ProcessGlobals, FileGlobals and SapiGlobals

joehoyle:
This is akin to ExecutorGlobals, but for the process globals


joehoyle:
@ptondereau do you know what config / version to build the bingings from? I think that's why this is failing!

ptondereau:
Hey @joehoyle if you want to update the doc binding, you need to use the latest stable version of PHP with debug mode. 
When 8.3 will be out, you should update it and use this one

danog:
Mind rebasing pls?

joehoyle:
@danog ah thanks, was also just fixing conflicts but looks like you got it! I was struggling to regenerate docsrs_bindings. I'm using homebrew php@8.3-debug` but it seems that's not actually a ZEND_DEBUG build of php

danog:
>I was struggling to regenerate docsrs_bindings

Yeah I just compiled php-src from source with `--enable-debug` and ran `cargo clean; cargo build; cp target/debug/build/ext-php-rs-*/out/bindings.rs docsrs_bindings.rs`, pretty simple

joehoyle:
Ah ok cool, yup got that now. However it looks like maybe the ones in the repo are generated from Linux? I'm seeing some additions for Darwin when I'm running it under macos

PR #253, opened 2023-07-19
Add StreamWrapper API

joehoyle:
This allows extensions to register and unregister custom stream wrappers

Still need to add docs.


danog:
Looks awesome, could you take a look at the clippy failures though? :)

joehoyle:
Ah this had `process-globals` as the upstream as it relied on some of that stuff. I think it might be better to revert this merge into `process-globals` as this is failing quite a few clippy checks. We can then put it in draft

danog:
I've already fixed everything, no need to revert

joehoyle:
Ok nice one

PR #256, opened 2023-07-19
Implement IntoZvalDyn for Zval

joehoyle:
When using call_user_func!() / Callable::try_call() it's currently not possible to use Zvals in the arguments which is unexepected. The simplest way is to add the trait to Zval's too.


ptondereau (approved):
Nice

PR #258, opened 2023-07-19
Add helper method for function table

joehoyle:


danog:
Not sure this is needed anymore, https://github.com/davidcole1340/ext-php-rs/pull/264/, already offers a Function abstraction with try_from_name methods.

joehoyle:
If you want to programmatically add and remove functions from the function table, you'd still need this

danog:
Oh, right!
Could you please rebase then :)

PR #261, opened 2023-07-19
Add helper to get the sapi name

joehoyle:
Useful to determine if the context is CLI / FPM etc


danog:
Hi, new maintainer of ext-php-rs here, mind rebasing&fixing the failing tests? :)

PR #268, opened 2023-10-17
feat(iterator): add helper for zend_object_iterator and iterable type

joelwurtz:
Not sure if this is wanted, so this is a first shot into adding a `ZendIterator` type to handle all objects that have an iterator and allow custom extension to use them in a more friendly way

It should work for any class that extends `Traversable`, but not for array, a super type may be needed if we want to support `iterable` type : which is mainly either an array or a Traversable

TODO: 

 * [x] documentation


src/types/iterator.rs
```diff
@@ -0,0 +1,106 @@
...
+        unsafe {
+            get_current_key(&mut *self, &mut key);
+        }
+
+        Some(key)
+    }
+}
+
+impl<'a> Iterator for Iter<'a> {
+    type Item = (u64, Option<String>, &'a Zval);
```
danog, line 64: Do you think we could turn the (u64, Option<String>) into an ArrayKey enum of u64/String?

src/types/object.rs
```diff
@@ -103,6 +103,17 @@ impl ZendObject {
         unsafe { self.ce.as_ref() }.expect("Could not retrieve class entry.")
     }
 
+    /// Returns the [`ClassEntry`] associated with this object.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the class entry is invalid.
+    pub fn get_class_entry_mut(&self) -> &'static mut ClassEntry {
```
danog, line 111: ?

danog (changes requested):
Thanks for the contribution!
Could you also add some helper methods to obtain a ZendIterator from a Zval?

src/types/iterator.rs
joelwurtz: I keep the same item type as the array iterator (so an extension may use either array and traversable and keep the same api) but if you prefer that i'm fine with it

src/types/object.rs
joelwurtz: I need a `& mut class entry` to get the iterator, not sure if this is best way to retrieve it otherwise i can directly get it from the raw pointer of the object

joelwurtz:
> Could you also add some helper methods to obtain a ZendIterator from a Zval?

It's possible however i'm not sure about method name as `get_iterator` may be mistaken to a class that implements the `Iterator` interface ? where in our case `ZendIterator` represents all object that have a `zend_object_iterator` which is more linked to the `Traversable` interface and even other php methods

src/types/iterator.rs
danog: Actually, I think it might be a nice idea to also change the array iterator api...

danog:
>It's possible however i'm not sure about method name as get_iterator may be mistaken to a class that implements the Iterator interface

`get_traversable` looks like a good idea, or even better zval could direclty impl IntoIterator.

src/types/iterator.rs
joelwurtz: I have changed both api to use an enum for key

joelwurtz:
I added `is_traversable`, `traversable`, `is_iterable`, `iterable` to the zval implementation

iterable will handle both array and traversable (like php does), i also added the `iterable` type so it will be correctly generated into stub.

I try to look into having `IntoInterator` for the Zval but i'm not sure if this is wanted, as not all zval can have an iterator and it also consume self, which means it need a need iter type keeping the zval and having an optional iterator, it's still possible but i don't think it's really needed ?

joelwurtz:
It should be ready for review, i added the documentation for iterator and iterable types

joelwurtz:
FYI, the new `IterKey` struct for consuming an array will be a BC break for existing extension using this library

joelwurtz:
I'm wondering also if it should better support exception ?

Actually if the user provide a generator for the iterator, the method cannot catch something if it throws (but php will correctly throw the exception at the end of the method call)

joelwurtz:
Not sure why the test are failing ?

danog:
>I'm wondering also if it should better support exception ?

Yep it absolutely should

ptondereau:
> Not sure why the test are failing ?

I think you didn't update the https://github.com/davidcole1340/ext-php-rs/blob/master/docsrs_bindings.rs files with the new bindings you've added

joelwurtz:
How do you update this file ? I try to copy what i add in the bindings.rs file but it seems it not the same content, and it seems that it was generated with debug to true (which change some structure and constant)

ptondereau:
> How do you update this file ? I try to copy what i add in the bindings.rs file but it seems it not the same content, and it was generated with debug to true (which change some structure and constant)
> 
> PS: I updated the file manually in the meantime

There is a note here: https://github.com/davidcole1340/ext-php-rs/blob/1a8211c392ea33a6192e4556b99762d704991cab/allowed_bindings.rs#L8-L20

We enforce debug mode while generating it because it exposes more data and structs (especially for the `emalloc/efree` part

danog (approved):

danog:
Changed my mind on the design a bit, refactoring...

danog:
Thanks!

Issue #99, opened 2021-10-05
Rust analyzer macro expansion errors


davidcole1340:
With this [change] in rust-analyzer, expansion of ext-php-rs macros errors out as they are no longer expanded in order. This is because Rust provides no guarantees of macro expansion, and the library abuses the fact that rustc expands them (at the moment) in order by maintaining a global proc-macro state.

Currently the solution to this is to disable macro expansion in rust-analyzer:

```
"rust-analyzer.experimental.procAttrMacros": false
```

It could be best to move to remove this global state. The global state is used to maintain a register of all the entities to be registered in the PHP module startup and get module functions.

[change]: https://github.com/rust-analyzer/rust-analyzer/pull/10366

PR #270, opened 2023-10-19
feat(embed): add embed features, add test example which run php inside it

joelwurtz:
This add a new features "embed" which links result with the `libphp.so` library available if php-embed is installed

This should allow in the future to test code directly in the wanted file, there is still a lot to do but this is a small PoC so we can iterate on that and have a better test suite


ptondereau (commented):
Amazing!

This PHP module should be compiled with the `./configure --enable-embed` but the CI is currently using a pre-compiled version of PHP from: https://github.com/shivammathur/setup-php,
 that why the CI is failing here

joelwurtz:
Not sure if enable embed is need as it's only to build the libphp.so file from what i understant.

And yes the CI is failing because it needs this shared object ant it's not available on https://github.com/shivammathur/setup-php,

I will do a PR on this repo to add php-embed, in the meantime we can avoid --all-features and run embed test with test on the Dockerfile ?

ptondereau:
> Not sure if enable embed is needed as it's only to build the libphp.so file from what i understand.
> 
> And yes the CI is failing because it needs this shared object ant it's not available on https://github.com/shivammathur/setup-php,
> 
> I will do a PR on this repo to add php-embed, in the meantime we can avoid --all-features and run embed test with test on the Dockerfile ?

Sounds good to me, but I'm curious that if the embed binding is included in the wrapper, does the feature flags really work here? I mean, for example, I don't want to build extra bindings that are only used in the test even if I've disabled the `embed` flag. Can you have a look?

joelwurtz:
Not sure i follow, from what i understand : 

 * You use embed if you want to create a binary that use php behind (like a new sapi, or testing this crate)
 * You don't use embed if you want to compile / build an extension
 
 I assume if you enable embed and use php_module it will not work / it would be dangerous as you would run php with a lib that use bindings from libphp (not sure which function the module will use, one from php ? one from libphp ?)
 
 Also didn't know that test were run in parallel inside rust, which means i have to force thread count to 1 when testing (or have php embed compiled with zts)

ptondereau:
IMHO, I don't want a developer who is building a rust extension with this lib, to have `libphp.so` to build its extension (even in the CI).

My first guest would be to create a new crate in the workspace with the required bindings + `libphp` with only test cases. But I'm not very confident and let, maybe @danog has a word on it.

> It would be dangerous as you would run php with a lib that use bindings from libphp (not sure which function the module will use, one from php ? one from libphp ?)

I think it will use `libphp` only when you wrap the code between `php_embed_init`/`php_embed_shutdown`

ptondereau:
> This will not be linked unless you have the "embed" feature specified

It's ok for me then, lezgo!

About the crate splitting, let's wait and discuss it in a separate issue, but TBH, I would love to split bindings + safe wrappers from the macro. The problem is: There is an initial refactor of the macro system that needs to be done (see: https://github.com/davidcole1340/ext-php-rs/issues/99)

joelwurtz:
I updated the CI so the embed will be build on a specific workflow (as it's not available with setup-php) 

I also added a guard around the embed as it's not thread safe (in nts), but thread safe version can be added if it's available, i'm not sure where this PR should stop, i think it should bring the basic setup for test and rest could be added in new PRs.

src/zend/embed.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        return EmbedWriteGuard {
+            guard: write_guard,
+            data: Ok(()),
+        };
+    }
+
+    // We use a write guard, this should only be called once at a time in non thread safe
+    // In the future we may not need this when it's linked to a thread safe php
+    pub fn run<'a>(code: &str) -> EmbedWriteGuard<'a, Result<Zval, EmbedError>> {
```
ptondereau, line 58:
```suggestion
    pub fn eval<'a>(code: &str) -> EmbedWriteGuard<'a, Result<Zval, EmbedError>> {
```

I would rename to `eval` just to be sure that end user understand what it does

ptondereau (approved):
LGTM! I think we have a good first iteration here

joelwurtz:
I dont think it's ready yet, i'm rewriting the logic for the embed as it's initialized only once, which means php code in some test can pollute other test

like a test setting a variable `$foo`, another test may use this variable latter which will be already defined

I'm actually rewriting the API so someone will do the following : 

```php
Embed::run_fn(|| {
    let result = Embed::eval("'foo';");

    assert!(result.is_ok());

    let zval = result.as_ref().unwrap();

    assert!(zval.is_string());
    assert_eq!(zval.string().unwrap(), "foo");
});
```

.github/actions/embed/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN apt update -y && apt upgrade -y
+RUN apt install lsb-release wget gnupg software-properties-common -y
+RUN bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
+
+ENV RUSTUP_HOME=/rust
+ENV CARGO_HOME=/cargo
+ENV PATH=/cargo/bin:/rust/bin:$PATH
+
+RUN (curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain nightly --no-modify-path) && rustup default nightly
```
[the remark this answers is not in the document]
joelwurtz: Not at all i will change it

joelwurtz:
Ready for last review

src/embed/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+            opened_path: null_mut(),
+            type_: 0,
+            primary_script: false,
+            in_list: false,
+            buf: null_mut(),
+            len: 0,
+        };
+
+        unsafe {
+            zend_stream_init_filename(&mut file_handle, path.as_ptr());
```
ptondereau, line 41: Just a question: why not use a `fopen` here instead of a stream?

src/embed/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+static RUN_FN_LOCK: RwLock<()> = const_rwlock(());
+
+impl Embed {
+    pub fn run_script<P: AsRef<Path>>(path: P) -> bool {
+        let path = CString::new(path.as_ref().to_str().unwrap()).unwrap();
+        let mut file_handle = zend_file_handle {
+            handle: _zend_file_handle__bindgen_ty_1 { fp: null_mut() },
+            filename: null_mut(),
+            opened_path: null_mut(),
+            type_: 0,
```
ptondereau, line 33: I think you can rely on `zend_stream_type` here

Cargo.toml
```diff
@@ -35,6 +35,7 @@ zip = "0.6"
 
 [features]
 closure = []
+embed = []
```
ptondereau, line 38: Just a reminder to put in the doc that this feature will only work on linux OS for now

src/embed/mod.rs
joelwurtz: I don't really care here about the value, it's correctly initialized when using the `zend_stream_init_filename` (in php code it initialize an empty structure without values)

src/embed/mod.rs
joelwurtz: Because php does a lot of background work to correctly initialize this structure, i think it will be overkill to try to reproduce what php does already

Cargo.toml
joelwurtz: Done

ptondereau (approved):
Thank you for the contribution!

PR #272, opened 2023-10-20
feat(embed): correctly handle panic inside embed

joelwurtz:
This correctly shutdown php runtime when there is a panic (like a failed test) inside the embed sapi, otherwise a panic would cause a segfault (since php runtime is not correctly shutdown)


ptondereau (approved):
Good addition thank you!

Issue #138, opened 2022-05-08
Constructors with subclasses


joehoyle:
I'm not sure it's possible for classes via the php_impl macro to be subclassed, as the constructor will return `Self`, instead of the subclassed class. I might be misunderstanding how subclassing (from user land PHP) should be working though!

joehoyle:
@joelwurtz I wonder if you've run into this issue / found a solution to this issue?

joehoyle:
or maybe @danog !

joehoyle:
I _think_ the issue is around here: https://github.com/davidcole1340/ext-php-rs/blob/dddc07f587cd5d40f3b3ff64b9a59fba8299d953/src/builders/class.rs#L164C58-L168 where PHP passes a reference to the `ce` (which would be the subclass), but ext-php-rs ignores the `ce` provided by PHP and instead takes the static reference for the `ce` from the registered class' metadata.

ju1ius:
@joehoyle Yes that's the issue. The `create_object` handler should just forward the `zend_class_entry` pointer to [internal_new](https://github.com/davidcole1340/ext-php-rs/blob/dddc07f587cd5d40f3b3ff64b9a59fba8299d953/src/types/class_object.rs#L105) (the Zend engine guarantees that it points to either your registered class entry or a subclass thereof).

Then this pointer should be passed to `zend_object_std_init` et al.

joehoyle:
@ju1ius ok cool, I had a quick go at trying that, but then ran into panics in the `free_obj` handler and such so I wasn't sure if I was going in the right direction or not!

ju1ius:
Yeah, I tried to fix it sometime ago but ran into so much fundamental issues that I ended up writing my own library from scratch. 🤣

joehoyle:
@ju1ius ah I see! Any chance that code is open sourced?

joehoyle:
Ok I think I got this working as it happens in #277. I wasn't that familiar with how the `ce` was linked to the Rust struct, so I wrote up a playthrough as I understood it. If that's valuable to anyone else in the future, I've pasted it below:

---

# Class Entry

In this library, when a struct is marked as `#[php_class]`, this implements the `RegisteredClass` trait on the struct. This trait adds methods:
  - `get_metadata() : &'static ClassMetadata`
  - `get_properties() : HashMap<String, Property>`

`ClassMetadata` stores the `ClassEntry` (`ce`) for the `#[php_class]` struct.

[...]

PR #277, opened 2023-10-24
Forward ClassEntry in create_object

joehoyle:
See #138

The `ce` should be taken from the `create_object`. From that point on, everything else can carry on just about normal. the only thing that needs to change is the sanity check to make sure the `ClassEntry` for the object passed to the object handlers is a child as well as the direct class.


joehoyle:
Ping @danog :)

PR #278, opened 2023-10-30
[Feature] custom SAPI

joelwurtz:
Goal of this PR is to provide a safe API around creating it's own SAPI for PHP

Actually there is only basic stuff and no wrapper at all and a test to show / confirm how it works.

I'm not sure about which API to provide to make things easier if someone want to create it's own SAPI.

Macro are out of scope for this PR (as i prefer to wait for the new macros to be merged and i think we can do a first version without them)


src/builders/sapi.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Sets the send header function for this SAPI
+    ///
+    /// # Arguments
+    ///
+    /// * `func` - The function to be called on shutdown.
+    pub fn send_header_function(mut self, func: SapiSendHeaderFunc) -> Self {
+        self.module.send_header = Some(func);
+        self
+    }
```
joelwurtz, line 67: a lot of thing can be defined but ATM this is the only required function for PHP to not segfault (that's why we set a dummy handler if nothing is set latter)

src/zend/try_catch.rs
```diff
@@ -26,14 +28,25 @@ pub(crate) unsafe extern "C" fn panic_wrapper<R, F: FnMut() -> R + RefUnwindSafe
 ///
 /// * `Ok(R)` - The result of the function
 /// * `Err(CatchError)` - A bailout occurred during the execution
-pub fn try_catch<R, F: FnMut() -> R + RefUnwindSafe>(func: F) -> Result<R, CatchError> {
+pub fn try_catch<R, F: FnMut() -> R + RefUnwindSafe>(
+    func: F,
+    first: bool,
```
joelwurtz, line 33: not sure about adding a parameter here, instead of having another function which should be more explicit

tests/sapi.rs
```diff
@@ -0,0 +1,81 @@
...
+    unsafe {
+        ext_php_rs_sapi_startup();
+    }
+
+    unsafe {
+        sapi_startup(sapi);
+    }
+
+    unsafe {
+        php_module_startup(sapi, module);
```
joelwurtz, line 31: here we create our own sapi with our own module so we can add function to php, this avoid the previous hack (zend_register_module) which does not work for a lot of cases (like registering classes)

tests/sapi.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    let result = unsafe { php_request_startup() };
```
joelwurtz, line 34: Since we separate module startup / request startup we can imagine having a loop here to handle multiple requests / execution with memory cleaned between them

src/zend/try_catch.rs
ptondereau: IMHO, I would create a new function that reflect the underlying zend api

src/builders/mod.rs
```diff
@@ -4,7 +4,11 @@
...
 mod function;
 mod module;
+#[cfg(feature = "embed")]
+mod sapi;
 
 pub use class::ClassBuilder;
 pub use function::FunctionBuilder;
 pub use module::ModuleBuilder;
+#[cfg(feature = "embed")]
+pub use sapi::SapiBuilder;
```
ptondereau, line 14: could it be its own feature decoupled from the `embed` one?

src/builders/sapi.rs
ptondereau: That's enough for me, let's make it evolve when users will have more usage

src/builders/mod.rs
joelwurtz: It should be possible be it's useless since you use SAPI when you build a binary, and in this case you need to be linked to libphp which is only available with the embed feature

PR #284, opened 2023-11-21
Add PHP 8.3, SAPI globals support

danog:


joehoyle:
Nice one, but why are we merging failing builds?

danog:
Only Windows builds were failing, due to some delay on PHP's part in the release of the 8.3 SDK; mac and linux builds were all passing, so a quick re-run of workflow once the SDK gets released will fix everything.

PR #288, opened 2023-11-24
feat(iterator): add helper for zend_object_iterator and iterable type

danog:


danog:
Oh nevermind I can push directly to that branch

Issue #290, opened 2023-11-24
Memory corruption when running module tests


danog:
```
[daniil@arch-lenovo ext-php-rs]$ cargo test --all-features --no-fail-fast --test module
    Finished test [unoptimized + debuginfo] target(s) in 0.25s
     Running tests/module.rs (target/debug/deps/module-d6deba67b910978c)

running 1 test
malloc_consolidate(): unaligned fastbin chunk detected
error: test failed, to rerun pass `--test module`

Caused by:
  process didn't exit successfully: `/home/daniil/repos/ext-php-rs/target/debug/deps/module-d6deba67b910978c` (signal: 6, SIGABRT: process abort signal)
error: 1 target failed:
```

Opening this more as a reminder to myself, as the CI tests don't run the module tests.

Xenira:
@danog was this fixed with #294? Tests seem to work right now.

PR #294, opened 2023-11-24
Add integration tests

danog:
Reverts davidcole1340/ext-php-rs#293


danog:
Don't have a mac os machine so can't easily debug the mac os failures, would be really happy if someone could give a try...

joehoyle:
Oh I see why it's only being triggered on the Debug build, it seems this check is only made for internal functions if ZEND_EDBUG:

```c
#if ZEND_DEBUG
static ZEND_COLD void zend_verify_internal_return_error(const zend_function *zf, zval *value)
{
	...

static ZEND_COLD void zend_verify_void_return_error(const zend_function *zf, const char *returned_msg, const char *returned_kind)
{
	...
}

static bool zend_verify_internal_return_type(zend_function *zf, zval *ret)
{
	...
}
#endif
```

joehoyle:
Yeah I'd imagine we should be running all the test envs with a debug build in that case. i think you may be able to reproduce if run you against a debug build on linux?

joehoyle:
FYI we are setting the PHP build to debug on macos here: https://github.com/davidcole1340/ext-php-rs/blob/master/.github/workflows/build.yml#L48

danog:
>Oh I see why it's only being triggered on the Debug build, it seems this check is only made for internal functions if ZEND_EDBUG:
> think you may be able to reproduce if run you against a debug build on linux?

Indeed, I encountered the exact same issue previously on a linux debug build, which lead to the discovery of a bug in ZendType (a simple C string was used instead of a zend string for the type name), which I fixed in https://github.com/davidcole1340/ext-php-rs/commit/86c9074470a8e1ea4742eadde026159ede7e7528, but then the same issue started occurring again here, only on Mac OS builds, and not on a local Linux debug build (which has different issues, namely https://github.com/davidcole1340/ext-php-rs/issues/290 + a segfault when running library tests probably due to memory corruption as for #290).

>FYI we are setting the PHP build to debug on macos here: https://github.com/davidcole1340/ext-php-rs/blob/master/.github/workflows/build.yml#L48

Actually shouldn't that enable debug builds for all OSes?

joehoyle:
Hmm ok I think https://github.com/davidcole1340/ext-php-rs/commit/86c9074470a8e1ea4742eadde026159ede7e7528#diff-758fcaf4f7ca5c65a01fb257423dc04af5d266f93547a89f2559bec5410b2f1aL85 might be the key. To also test this I can do:

```php
$r = new ReflectionFunction('test_class');
var_dump( $r->getReturnType()->getName() );
```

That gives me `string(1) ""` which I think is the corrupt. When I change it back to a CString `ptr: CString::new(class_name).ok()?.into_raw() as *mut c_void,` then I get the correct return type:

```php
string(9) "TestClass"
```

danog:
The issue was caused by the fact that php 8.2 and below want a C string in the arginfo to then convert it internally to a zend string, but php 8.3 wants a C string only if the literal flag is set, and a zend string otherwise.

PR #296, opened 2023-11-28
Pub ArrayKey and update example

joehoyle:
Currently it's not possible to match against the array key, as it's not exported from mod.rs it seems.


danog:
My bad, thanks!

PR #298, opened 2023-11-28
Remove type coercion

danog:


joehoyle:
@danog I just discovered this when upgrading to 0.12.0. This seems like a kinda major BC break as it effectively makes all `#[php_function]`s now strictly typed. Do we have a recommended way to handle type coercion ala PHP non-scrict behaviour?

Xenira:
A question would be if we even want to support this.

I do not like that php does this and if documented I would be fine with sticking to strict types.

joehoyle:
I think it's probably a better default indeed, PHP type coercion is also context-dependant I believe too. Perhaps we _could_ have `zval.try_into_type()` or something where we could have some convenience methods.

danog:
Yeah this is a BC break, which is why I released a major.
I believe it's better to default to strict_types behavior, and I don't think it's a good idea to provide coercion methods, as we can coerce using directly rust...

PR #301, opened 2023-12-01
Try tests with Cstring

joehoyle:


danog:
Ahh there's another `_ZEND_TYPE_LITERAL_NAME_BIT` flag, used to signify that type.ptr is not a `zend_string*` but a `const char*`

joehoyle:
Ahhh! Nice find :D

joehoyle:
I think this might be different across PHP versions.. or something. IN PHP 8.3, there is:

```
/* Only one of these bits may be set. */
#define _ZEND_TYPE_NAME_BIT (1u << 24)
// Used to signify that type.ptr is not a `zend_string*` but a `const char*`,
#define _ZEND_TYPE_LITERAL_NAME_BIT (1u << 23)
```

whereas PHP 8.0:

```
#define _ZEND_TYPE_NAME_BIT (1u << 23)
```

PR #302, opened 2023-12-05
Add support for Enums

joehoyle:
This adds an EnumBuilder to create PHP enums, and a `#[php_enum]` macro which you can attach to Rust enums.

Only symbolic enums are supported from Rust. PHP doesn't support enums that hold data. They do support scalar backend-enums, however Rust does not support that. So, the `#[php_enum]` can only be used for non-backed enums. You can implement backed enums via the EnumBuilder in cses where you'd want to implement your own IntoZval / FromZval for your name and do something custom with the backed value.

We could potentially add attributes to the enum for their string / int scale values in PHP though.

I've added cfg php8.1 to this, as enums are only supported in PHP >= 8.1. I think it's better to conintue supporting PHP 8.0 for the time being, and just have the enums only be available in >= php 8.1.


kakserpom:
@joehoyle PHP's backed enums are totally supported in Rust ;-) 
```
enum Fruit {
    Apple,
    Pear,
    Orange,
}
impl Display for Fruit {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Fruit::Apple => write!(f, "A"),
            Fruit::Pear => write!(f, "P"),
            Fruit::Orange => write!(f, "O"),
        }
    }
}
impl TryFrom<String> for Fruit {
    type Error = String;
    fn try_from(s: String) -> Result<Self, Self::Error> {
        match s.trim() {
            "A" => Ok(Fruit::Apple),
            "P" => Ok(Fruit::Pear),
            "O" => Ok(Fruit::Orange),
            other => Err(format!("Invalid fruit: {}", other)),
        }
    }
}
```

Or just https://docs.rs/strum/latest/strum/derive.EnumString.html

kakserpom:
@Xenira what do you think about this PR? It it salvageable?

Xenira:
@kakserpom maybe. Want to finish next release first though. But enums are high on the list of stuff to do next.

Xenira:
Had a look and will rebuild this using changes from this MR, as that will probably be easier. A lot of the macro stuff has changed.

Xenira:
Implemented in #489

PR #315, opened 2024-05-25
feat(bindings): add bindings for cached function calls

Xenira:
For my extension I need to perform a lot of callbacks.

Using fci (fcall_info) and fcc (fcall_info_cache) yielded ~10% performance improvement.

If I find the time Ill port it into this lib. For now I just added the bindings to be able to use them in my code directly.

You might need to check the `docrs_bindings.rs` as it seems there are more changes then there should be (maybe php version?). Mainly `u8` -> `u16`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,7 @@ jobs:
           key: ${{ matrix.os }}-llvm-${{ matrix.clang }}
       - name: Setup LLVM & Clang
         id: clang
-        uses: KyleMayes/install-llvm-action@v1
+        uses: KyleMayes/install-llvm-action@v2
         if: "!contains(matrix.os, 'windows')"
         with:
           version: ${{ matrix.clang }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,6 +75,7 @@ jobs:
         with:
           version: ${{ matrix.clang }}
           directory: ${{ runner.temp }}/llvm-${{ matrix.clang }}
+          arch: ${{ matrix.os == 'macos-latest' && 'arm64' || 'x64' }}
           cached: ${{ steps.cache-llvm.outputs.cache-hit }}
       - name: Configure Clang
         if: "!contains(matrix.os, 'windows')"
--- a/allowed_bindings.rs
+++ b/allowed_bindings.rs
@@ -56,6 +56,8 @@ bind! {
     std_object_handlers,
     zend_array_destroy,
     zend_array_dup,
+    zend_call_function,
+    zend_fcall_info_init,
     zend_call_known_function,
     zend_fetch_function_str,
     zend_hash_str_find_ptr_lc,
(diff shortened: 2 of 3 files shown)


Xenira:
Maybe specifying the arch helps. Just trying stuff though.

Xenira:
Ummm, `arch` seems to be only supported from v2. Updated the action.

Xenira:
Seems to still have x86_64 clang. `install-llvm-action` seems to be using cache. Can we try to clear the cache?

Xenira:
How do we proceed with the pipeline issue?

ptondereau:
BTW, if you want to add more bindings, you need to have the most up-to-date stable version of PHP (8.3.7 for now) with debug mode

Xenira:
> Hey @Xenira Thank you for your patience, I've cleaned the cache and re-runned the CI and now, you have another error.

Thanks! Did a quick google search and it seems to be a problem with dylib and the mac libc https://discourse.llvm.org/t/apples-libc-now-provides-std-type-descriptor-t-functionality-not-found-in-upstream-libc/73881

Ill try to set up my own pipeline to iterate faster on this.


> BTW, if you want to add more bindings, you need to have the most up-to-date stable version of PHP (8.3.7 for now) with debug mode (it exposes more symbols for bindgen)

I'm on 8.3.7, not sure about debug though. Is that related to the differences in `docrs_bindgen.rs`? Because the changes itself worked fine and I was able to use the new bindings.

ptondereau:
> Hey @ptondereau,
> 
> Finally got to test it.
> 
> This seems to be caused by clang / llvm 15 and 17. Supposedly it should work with 18, but there is no macos build yet.
> 
> I managed to get it to work by skipping the clang / llvm setup entirely for macos https://github.com/Xenira/ext-php-rs-pipeline-debug/actions/runs/10254367953/job/28368987959
> 
> How should we proceed with this? Easiest would be to skip the llvm setup for the macos runs.

Yes I've tried to get a full build on macos too but as you said, 18 is not available...
I'm not a huge fan about skipping MacOS for the moment but maybe @danog has some plan?

Xenira:
Wouldn't skip macos entirely. Just leaving out the setup steps makes macos build again https://github.com/Xenira/ext-php-rs-pipeline-debug/blob/0298ee6a016d1ddcb7178e904601485731746723/.github/workflows/build.yml#L64

ptondereau:
> Wouldn't skip macos entirely. Just leaving out the setup steps makes macos build again https://github.com/Xenira/ext-php-rs-pipeline-debug/blob/0298ee6a016d1ddcb7178e904601485731746723/.github/workflows/build.yml#L64

Ah ok! I'm ok about that, just leave a comment about the why and it should be good for me

Xenira:
Well, now clippy complains all over the place. Will create a seperate mr to get pipeline fixed.

Xenira:
Closing this and will have a look at adding cached fn directly.

This is covered by #406 as well

PR #320, opened 2024-08-07
ci: fix pipeline

Xenira:
Fixed ci as discussed in https://github.com/davidcole1340/ext-php-rs/pull/315

Also had to fix clippy rules that changed and made master fail.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,27 +64,26 @@ jobs:
       - name: Cache LLVM and Clang
         id: cache-llvm
         uses: actions/cache@v3
-        if: "!contains(matrix.os, 'windows')"
+        # Macos build doesn't work with clang < 18. As a build for version 18 is not available, we skip the setup
+        if: "!contains(matrix.os, 'windows') && !contains(matrix.os, 'macos')"
         with:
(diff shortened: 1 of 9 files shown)


ptondereau (approved):
LGTM for me maybe an extra eye from @danog ?

danog:
Not a super fan of removing platform tests tbh, I see that no one has contributed a mac build for clang 18 to the official llvm repo, but maybe there's some source that could be used for the binaries? Don't have a Mac, but it seems brew offers a clang 18 build...

Xenira:
Just added a `clang --version` to this and it seems to build with `Apple clang version 15.0.0 (clang-1500.3.9.4)` (https://github.com/Xenira/ext-php-rs-pipeline-debug/actions/runs/10329884456/job/28598214318#step:10:15).

Adding clang 18 is something I would do in #318.

Apple builds itself are not skipped, They just use the builtin clang, which is the better test imo as that better replicates a user trying to build on their system.

Xenira:
Hey, would like to get this moving again. Imo we should get the ci working again with as few changes as possible and then focus on getting clang 18 support.
@danog

PR #323, opened 2024-09-18
fix(linter): address linting error blocking the CI

jdrouet:


ptondereau:
Hello and thanks for the PR.
This is already addressed here: https://github.com/davidcole1340/ext-php-rs/pull/320

And we don't want to disable the entire job for the moment

PR #330, opened 2024-10-21
Don't use symbolic links for git.

faassen:
When I ran `git status` I got this message:

```
warning: unable to access 'crates/macros/.gitignore': Too many levels of symbolic links
warning: unable to access 'crates/cli/.gitignore': Too many levels of symbolic links
```

Since git 2.23 `.gitignore` that is a symbolic link is ignored. I made them explicit now (though I'm not sure they're even needed)

### DIFF
--- a/crates/cli/.gitignore
+++ b/crates/cli/.gitignore
@@ -1,1 +1,6 @@
-../../.gitignore
+/target
+Cargo.lock
+/.vscode
+/.idea
+/tmp
+expand.rs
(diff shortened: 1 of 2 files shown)


Xenira:
Looks fine to me. Might need some refinement later on, but should solve the issue.

Would like to fix CI before merging though.

Xenira (approved):
Looks fine to me. Might need some refinement later on, but should solve the issue.

Would like to fix CI before merging though.

Xenira:
Rebased from master. Pipelines should succeed now.

Issue #331, opened 2024-10-21
`TRACK_VARS_REQUEST` seems to not be stored in `http_globals`


Xenira:
Rust nightly build detects that the access in https://github.com/davidcole1340/ext-php-rs/blob/0d9496b76cce8914469b60f90983283c2fdb0db0/src/zend/globals.rs#L287 results in a out of bounds.

Looking at the C code it seems that `TRACK_VARS_REQUEST` is not referenced anywhere.

Maybe someone with more knowledge about php internals can point towards where to find the `$_REQUEST` super global.

ptondereau:
Very weird as this constant lives here: https://github.com/php/php-src/blob/master/main/php_globals.h#L46

Xenira:
Ye, found the global both in that file and in the zend code https://github.com/php/php-src/blob/50acf5eb15df4be623e01af7926688ce8fc7529b/main/php_globals.h#L46

But looking at the array size (https://github.com/php/php-src/blob/50acf5eb15df4be623e01af7926688ce8fc7529b/main/php_globals.h#L114) it is `6` and therefore out of bounds with index `6`

ptondereau:
Ok but it seems that is related to Rust according to this issue: https://github.com/rust-lang/rust/issues/90534

Xenira:
Sure, might be my limited c knowledge, but shouldn't the `http_globals` be constant length?

ptondereau:
So as we know that the 6th index exists, we could tell the compiler to disable this warning on this specific line instead

Xenira:
Just created a simple test:
```rust
#![cfg_attr(windows, feature(abi_vectorcall))]
use ext_php_rs::prelude::*;
use ext_php_rs::zend::ProcessGlobals;

#[php_function]
pub fn hello_world(name: &str) -> String {
    format!("Hello, {:?}!", ProcessGlobals::get().http_request_vars())
}

#[php_module]
pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
    module
}
```

Results in
```
thread '<unnamed>' panicked at /home/rippi/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ext-php-rs-0.12.0/src/zend/globals.rs:287:9:
index out of bounds: the len is 6 but the index is 6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
[...]

PR #332, opened 2024-10-21
fix(globals)!: disabled `$_REQUEST` super global function

Xenira:
Just a test for now. Goal is fixing CI. If merged with these changes:

BREAKING CHANGE: If you used `http_request_vars()` before it will now panic until a proper implementation is found.

Refs: #331


ptondereau:
I'm ok to merge it for now, as we're waiting for the maintainers answer.
Some linting issue BTW

Xenira:
Ye, wanted to wait for some info in the issue before putting in more work. Seeing the problematic pipelines pass was enough for a first throw.

Will add links to the issue here and on php side as well.

Xenira:
@ptondereau fixed the issue and added docs

Issue #329, opened 2024-10-21
explicit registration mode


faassen:
Since in #327 (and maybe #328) I run into issues with registration order, I think that Rust compilation order, which ext-php-rs relied has been broken at least under some circumstances, which screws up at the very least rust analyzer.

I started investigating the cause of this error. `ext-php-rs` has a static STATE that things get registered in during macro expansion.  This has lines like:

```rust
        if state.built_module {
            bail!("The `#[php_module]` macro must be called last to ensure functions and classes are registered.");
        }

        if state.startup_function.is_some() {
            bail!("The `#[php_startup]` macro must be called after all the classes have been defined.");
        }
```

which are causing the problem.

## Not checking

As a relatively simple hack we could just disable these lines. That would fix the macro errors I think, but then you'd get absolutely no warning that your functions and classes aren't actually being registered. That's risky as we are already getting errors about the order of registration, and the order of macro expansion is not guaranteed.

## Explicit mode

I started investigating what it would be required to have explicit registration. I envision a mode where you write

`#[php_class(explicit)` 

and you're then required to register it explicitly in the module builder.

What does explicit do?

With an `explicit` flag, we could instead generate static annotations for the various macros (such as a class description, or function description) that only later is explicitly referred to during registration.

Then you'd register things explicitly:

```rust
#[php_class(explicit)]
struct Foo {
}

#[php_impl(explicit)]
impl Foo {
}

#[php_module(explicit)]
pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
     php_class_explicit!(module, Foo)
}
```

I'll note that `php_class_explicit` is a macro so that it looks for associated registrations that are generated by the `php_class`. These registrations are stored by the macro in a module that's generated associated with `Foo`. Unfortunately we can't call this module `Foo` as we need access to the original `Foo` in non-macro code, so some magic will need to remain (you can't import Foo and register it in another module, you're better off implementing a new `get_module` in that module and then invoke it from the original one.

## Implementation challenges

Unfortunately the current system relies heavily on the ability to register things during compile time. It builds extensive structs during compile time (such as `Class`) which are then registered. Ideally in the explicit system we don't want such a global object at all, as in that case we'd still be dependent on import order which can break - `php_module` needs to be called at the very last for it to work, and we can't guarantee that.

[...]

Issue #327, opened 2024-10-16
Spurious macro errors during development


faassen:
During development in VS Code with Rust analyzer, even if the build succeeds, I still get a lot of spurious errors in my editor:

`"The `#[php_module]` macro must be called last to ensure functions and classes are registered.`

`Impls must be declared before you declare your startup function and module function`

These go away once I restart Rust analyzer, but appear again as soon as I make any edits. This makes for an uncomfortable development experience, especially given that the build options ext-php-rs requires also seem to make cargo recompile a lot more than usual.

Does anyone else see these errors? Is there a way to fix this? I assume something is going wrong with the order in which macros are executed, and that somehow the `#[php_module]` macro is executed too early under some conditions.

Xenira:
Guess it has to do with this:

> These macros do abuse the fact that (at the moment) proc macro expansion seems to happen orderly, on one single thread. It has been stated many times that this order is undefined behaviour ([see here](https://github.com/rust-lang/reference/issues/578)), so these macros could break at any time with a rustc update (let's just keep our fingers crossed).
>
> The macros abuse this fact by storing a global state, which stores information about all the constants, functions, methods and classes you have registered throughout your crate. It is then read out of the state in the function tagged with the #[php_module] attribute. This is why this function must be the last function in your crate.
>
> In the case the ordering does change (or we find out that it already was not in order), the most likely solution will be having to register your PHP exports manually inside the #[php_module] function.

https://davidcole1340.github.io/ext-php-rs/macros/index.html

Have the same error, but only sometimes. A little luck is involved getting stuff to execute in the correct order (or thats my guess at least).

Xenira:
I am using rust analyzer as well.
Also seems like it has gotten worse.

Both fixing the magic and a no magic way would be great.
Will have a look after work.

faassen:
Besides analyzer errors I am now also getting errors like this:

`PHP Fatal error:  Xee\SequenceIterator must be registered before Xee\Sequence in Unknown on line 0`

I'm not sure whether it's related. Yesterday when I got it I thought I fixed it by moving the SequenceIterator registration above the Sequence definition, but today with additional changes, I'm getting this error more consistently. If I can't fix this reliably and this is indeed due to auto-registration order issues then this is a bigger problem than just the analyzer being flaky. I'm going to do some digging.

PR #333, opened 2024-10-22
Explicit preparation phase without STATE

faassen:
State (through the static `STATE`) is modified and used throughout ext-php-rs, but it's giving rise to various problems during development - the order in which the run compiler runs things is not guaranteed, and it appears that at least in some cases (in rust analyzer?) the system gets confused. See #327. 

To make this go away, I want to introduce an explicit registration mode. This is a preparatory refactor (which should introduce no behavioral changes) to help go into this direction, unweaving STATE access and modification from generation of token streams. It doesn't do anything by itself, but hopefully clears the ground.

The next step would be to see whether we can defer building and registration to a later phase. In particular, the `ClassBuilder` `build` method registers as a side-effect, though the other builders don't appear to do so.

This is a step in the direction of #329

### DIFF
--- a/crates/cli/.gitignore
+++ b/crates/cli/.gitignore
@@ -1,1 +1,6 @@
-../../.gitignore
+/target
+Cargo.lock
+/.vscode
+/.idea
+/tmp
+expand.rs
--- a/crates/macros/.gitignore
+++ b/crates/macros/.gitignore
@@ -1,1 +1,6 @@
-../../.gitignore
+/target
+Cargo.lock
+/.vscode
+/.idea
+/tmp
+expand.rs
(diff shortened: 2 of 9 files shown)


crates/cli/.gitignore
Xenira, line 1: Can you check if this mr builds on top of current master. A little confused seeing these changes again.

Xenira:
@faassen Gonna close this as macros have changed entirely with #372.

Please let me know if something from this MR would be an improvement over the current impl.

Never the less, thanks for putting in the effort!

faassen:
Makes sense! I am going to take a look soon!

PR #334, opened 2024-10-22
fix(globals): correctly fetch `$_REQUEST` super global

Xenira:
Fixes the `http_request_vars(&self)` method as suggested in https://github.com/php/php-src/issues/16541#issuecomment-2427599561.

Refs: #331


Xenira:
@joehoyle can you review this? Would like to get this merged.

build.rs
```diff
@@ -221,9 +221,6 @@ fn check_php_version(info: &PHPInfo) -> Result<()> {
     // introduced in PHP 8.1).
     //
     // PHP 8.0 is the baseline - no feature flags will be introduced here.
-    //
-    // The PHP version cfg flags should also stack - if you compile on PHP 8.2 you
-    // should get both the `php81` and `php82` flags.
```
joehoyle, line 226: Did you need to change this as part of this PR? This is quite a but change as it seems maybe you're changing the fall0through beahviour of `php81` being set on `php82` etc?

docsrs_bindings.rs
```diff
@@ -2011,105 +2092,122 @@ pub struct _php_stream {
 }
 impl _php_stream {
     #[inline]
-    pub fn is_persistent(&self) -> u8 {
-        unsafe { ::std::mem::transmute(self._bitfield_1.get(0usize, 1u8) as u8) }
+    pub fn is_persistent(&self) -> u16 {
```
joehoyle, line 2095: Are all the changes to the bindings intentional? This looks like maybe you didn't compile it on the correct PHP version / system.

joehoyle (commented):
@Xenira there's quite a lot fo changes here whereby it's not clear that all are part of this PR. For example, all the lifetime changes ?

Xenira:
Thanks for looking at it.

> For example, all the lifetime changes ?

Those are because of new clippy rules. I could move them to a separate mr.

build.rs
Xenira: I needed this to differentiate `php80` and `php81`. Guess i could have done a `and(php80, not(any(php81, php82, php83, php84)))` but that seemed a little unwieldy.

But I can pull this into a separate mr.

joehoyle: yeah, I'd split out. I'd rather not making that change as part of this PR unless strictly necessary

docsrs_bindings.rs
Xenira: Hmmm, I used php 8.3 build from source with debug enabled on a 64bit machine.

joehoyle: Ah ok, I guess that got updated in https://github.com/davidcole1340/ext-php-rs/commit/f621e60be66425d3bf3ac1a5ebccf675cdda263c

build.rs
Xenira: #345

joehoyle (approved):

PR #338, opened 2024-11-21
PHP 8.4

joehoyle:


joehoyle:
It seems on windows, https://windows.php.net/download/ only has php 8.4.1 devel package, but currently the latest version when installing php via https://github.com/shivammathur/setup-php is 8.4.0

danog:
8.4.1 is the first official release, 8.4.0 was actually skipped due to security bugs

joehoyle:
@danog I wonder if 8.4.0 still got distributed some places? CI is showing:

> √ PHP Installed PHP 8.4.0-dev (d0c336a6663b924e6d71d3227988f8a6666f330e)

build.rs
```diff
@@ -245,6 +249,10 @@ fn check_php_version(info: &PHPInfo) -> Result<()> {
         println!("cargo:rustc-cfg=php83");
     }
 
+    if version >= PHP_84_API_VER {
+        println!("cargo:rustc-cfg=php84");
+    }
+
```
Xenira, line 255: in #334 i touched this as well, as having the cfg's set up this way didn't allow to detect specific versions. I think we should have a cfg only set if it is exactly that version of php.

https://github.com/davidcole1340/ext-php-rs/pull/334/files#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42R243

Xenira:
Regarding the version on windows we have two options here:

1. Check back periodically / contribute to the workflow and merge once its available
2.  Disable 8.4 on windows for now and have a follow up ticket.

I would be fine to go with 2 for now.

build.rs
joehoyle: Yeah, I think in this case it's actually what we want, as all future versions of PHP will want the changes in this PR, however agree that these fgs are a bit misleading as "php82" means "8.2 and higher"

Xenira: true

Xenira:
god, that solution is so stupid, but nice find

joehoyle:
I opened to https://github.com/shivammathur/setup-php/issues/885 see if there's any assistance

Xenira:
> I opened to [shivammathur/setup-php#885](https://github.com/shivammathur/setup-php/issues/885) see if there's any assistance

From what I can see this might actually be an issue in https://github.com/mlocati/powershell-phpmanager. But currently no windows available to test this. (Also powershell :vomiting_face:)

Xenira:
Now the correct version is downloaded, but there is no DevelPack for 8.4.1
https://windows.php.net/downloads/releases/archives/

Xenira:
There is one in https://windows.php.net/downloads/releases/ though

joehoyle:
Doh! Are those managed by Microsoft?

Xenira:
Don't know. Wasn't able to find information on where to report this.

Maybe just try finding it in releases if archives does not exist?

joehoyle:
Phew, we're passing finally!

joehoyle:
It's not clear what logic we should use for /archive or not. You're right that it might be best to check both, as I guess they might move at some point.

Xenira:
> It's not clear what logic we should use for /archive or not.

I would actually try release first as that should contain the main releases, and archive if not found.

Xenira (approved):
Looks good to me. Improvements in windows build can/should be separate issue.

joehoyle:
Shall I do a release 0.12.1? I don't think we're breaking anything so just a patch?

Xenira:
Tbh not long enough in the project to tell if we had breaking changes since last release.

Remember the `$_REQUEST` super global that now panics 'gracefully' but that never worked.

I would like to have a release bot like [release-plz](https://github.com/release-plz/release-plz) with proper release mr's to have a better overview over the changes.

Xenira:
@joehoyle did an automatic scan and it seems to be api compatible. Generated Changelog:

### Fixed

- *(globals)* [**breaking**] disabled `$_REQUEST` super global function ([#332](https://github.com/davidcole1340/ext-php-rs/pull/332))

### Other

- *(clippy)* fix new clippy errors
- Windows build should try archives
- fmt
- Use vs17 on php 8.4+
- Don't use archive for 8.4.1
- fmt
- Support php 8.4 internal api changes
- PHP 8.4
- fix pipeline ([#320](https://github.com/davidcole1340/ext-php-rs/pull/320))
- Update README.md ([#317](https://github.com/davidcole1340/ext-php-rs/pull/317))

pandanotabear:
When new release happen btw?

PR #339, opened 2024-11-26
Windows build should try archives

joehoyle:
Follow up to https://github.com/davidcole1340/ext-php-rs/pull/338

Windows bulds should check the releases page, and the archive, as it's not clear when they'll be moved by Microsoft.


Xenira:
LGTM

PR #345, opened 2024-11-28
refactor(cfg): use distinct cfg for php version

Xenira:
Refs: #334


joehoyle:
Thanks for splitting this out. Now we have a distinct PR, could you explain the motivation and advantages to doing this? This pattern was presumably introduced so you could be additive when checking for versions, e.g. "If we're on PHP 8.1, use a u16". The implication being that those kind of changes are very likely to be "PHP 8.1+" rather than just PHP 8.1.

I see you are instead recommending the pattern of "exclude all versions that wouldn't have that change" (e.g. not(any(7.4,8.0)). It's not really clear to me why this is any better?

PR #350, opened 2024-12-13
Windows - add a DEBUG build with debugging symbols

EdmondDantes:
This change enables the extension to be compiled correctly in debug mode.
The issue was that, in the default case, the builder attempted to use files from the PHP DEV PACK, which lacks debug symbols. This resulted in a linker error, as it couldn't find the emalloc functions (these have different signatures in DEBUG builds).

Solution:
* Add an environment variable `PHP_LIB`.
* Check if we are compiling in debug mode. If so, require this variable to be set. Otherwise, proceed as before.
* `PHP_LIB` must point to a built `PHP` instance. The library name is chosen as `php8ts_debug.lib`.

P.S. I tried to find a place in the documentation to add this, but I can't figure out where it should go. If you can provide details, I would certainly include it.

### DIFF
--- a/windows_build.rs
+++ b/windows_build.rs
@@ -234,13 +234,15 @@ impl DevelPack {
 
     /// Returns the path of the PHP library containing symbols for linking.
     pub fn php_lib(&self, is_debug: bool) -> PathBuf {
-
         let php_lib_path = std::env::var("PHP_LIB")
             .map(PathBuf::from)
             .unwrap_or_else(|_| self.0.join("lib"));
 
         if !php_lib_path.exists() {
-            panic!("Error: Specified PHP library path '{}' does not exist.", php_lib_path.display());
+            panic!(
+                "Error: Specified PHP library path '{}' does not exist.",
+                php_lib_path.display()
+            );
         }
 
         let candidates = if is_debug {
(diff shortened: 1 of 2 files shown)


Xenira:
Thanks for the contribution and sorry for the delay in getting back to you.

Regarding the documentation it would probably be best to add a separate "Windows" section here as I could not find a good place to put it.

Seems like the pipeline builds are now unable to find the proper php lib. Can you have a look at that?

EdmondDantes:
@Xenira Thank you for the suggestions.

I fixed the library search function to ensure it works in all cases or returns an error.
I added a brief note that for DEBUG builds on Windows, PHP must be manually built from the SDK.

windows_build.rs
```diff
@@ -233,13 +233,42 @@ impl DevelPack {
...
+To build the application in DEBUG mode on Windows,
+you must have a PHP SDK built with the DEBUG option enabled
+and specify the PHP_LIB to the folder containing the lib files.
+For example: set PHP_LIB=C:\php-sdk\php-dev\vc16\x64\php-8.3.13-src\x64\Debug_TS."#,
+                            php_lib_path.display()
+                        )
+                    } else {
+                        format!("Error: No suitable PHP library found in '{}'.", php_lib_path.display())
+                    }
+                ))
```
Xenira, line 271: use `expect` here?

EdmondDantes:
@Xenira Thanks!

Xenira:
@EdmondDantes could you run `cargo fmt` on your changes?

EdmondDantes:
@Xenira Done

Xenira:
Updated the branch with clippy fixes from master #352

Issue #343, opened 2024-11-28
Drop support for PHP 8.0?


Xenira:
PHP 8.0 is [end of life](https://www.php.net/supported-versions.php).

Should we drop support for it?

Xenira:
Deprecation warning was added. Will create a follow up for removing CI and slight adjustment to warning message during build.

PR #353, opened 2025-01-09
chore(php): add deprecation warning for php 8.0

Xenira:
Refs: #343


Xenira:
Fair point. Will change the warning message and docs accordingly.

Xenira:
@joehoyle like this?

joehoyle (approved):

PR #355, opened 2025-01-28
Add support for empty immutable shared arrays

mszabo-wikia:
Why:

- Since PHP 7.3, it's possible for extensions to create zvals backed by an immutable shared hashtable via the `ZVAL_EMPTY_ARRAY` macro.
- This helps avoid redundant hashtable allocations when returning empty arrays back to userland PHP code, and could likewise be beneficial for Rust extensions too.

What:

- Add `ZendHashTable::new_empty_immutable()` to obtain a `ZendHashTable` that is actually an empty immutable shared hashtable.
- Add `ZendHashTable::is_immutable()`. Use it to avoid attempting to free the immutable shared hashtable on drop, and to set appropriate type flags when initializing a zval with a `ZendHashTable`.
- Make `ZendHashTable`'s `TryFrom` implementations from `Vec` and `HashMap` return an empty immutable shared hashtable if the input collection was empty. Although this would allow every user to automatically benefit from this optimization, I'm not convinced about this part because this is technically a breaking change for consumers that construct `ZendHashTable`s via these converters in their Rust code. Maybe it'd be better to not change the converters and let projects explicitly use `::new_empty_immutable` if they deem the optimization would be useful.


Xenira:
@mszabo-wikia Thank you for the contribution. I have not that much time, but will try to have a look soon.

From a quick look `docsrs_bindings.rs` needs to be updated.

[PHP 8.2 also seems to fail.](https://github.com/davidcole1340/ext-php-rs/actions/runs/13005205850/job/36433597066?pr=355#step:15:91)

mszabo-wikia:
Thanks, I've regenerated the bindings using a PHP 8.2 debug build. It seems as though the bindings might not have been updated in a while, as there were some other unrelated changes. Or do they need to be generated using 8.0?

Xenira:
@mszabo-wikia pretty sure its php 8.3, but no one seems to know https://github.com/davidcole1340/ext-php-rs/pull/334#discussion_r1862652167

mszabo-wikia:
@Xenira ah thanks, this is valuable context. I've regenerated the bindings using PHP 8.3.

Xenira:
> Make ZendHashTable's TryFrom implementations from Vec and HashMap return an empty immutable shared hashtable if the input collection was empty

I think we should not do this, as returning a immutable hashmap is not the expected behaviour here

mszabo-wikia:
> > Make ZendHashTable's TryFrom implementations from Vec and HashMap return an empty immutable shared hashtable if the input collection was empty
> 
> I think we should not do this, as returning a immutable hashtable is not the expected behaviour here

Yeah, I had doubts about that myself. I've removed this logic.

Xenira:
@mszabo-wikia fixed the issue in nightly tests (#357). Could you rebase onto master.

You also need to update your commit message, as the new pipeline enforces conventional commits.

mszabo-wikia:
Perfect, thanks—I was wondering about that since I didn't experience the same error locally, but I have stable Rust installed which explains it.

src/types/array.rs
```diff
@@ -71,6 +71,27 @@ impl ZendHashTable {
...
+    /// ```
+    pub fn new_empty_immutable() -> ZBox<Self> {
+        unsafe {
+            // SAFETY: zend_empty_array is a static global.
+            let ptr = (&zend_empty_array as *const ZendHashTable) as *mut ZendHashTable;
+
+            // SAFETY: `as_mut()` panics if the pointer is null.
+            ZBox::from_raw(ptr.as_mut().expect("zend_empty_array inconsistent"))
+        }
+    }
```
Xenira, line 93: This should probably not return a owned vallue, as that allows mutability.

```rs
    let mut ht = ZendHashTable::new_empty_immutable();
    ht.insert("key", 1).unwrap();
```
would be valid but results in `signal: 11 (SIGSEGV) (core dumped)`

Xenira (changes requested):

src/types/array.rs
mszabo-wikia: Sorry for the delayed update—yeah, this definitely isn't ideal.

`std::ptr` only supports mutable pointers, so making the backing ptr immutable wouldn't play well with ZBox.

Maybe the right way to go here is to have a separate helper similar to the `RETURN_EMPTY_ARRAY` C macro in the Zend API, since really the only major use case for this in extensions is to return an empty array to userland without overhead. Then particularly performance-sensitive extensions can opt into using this as needed.

Xenira: Maybe just have a `struct ZendEmptyArray;` that implements `IntoZval`?

Xenira:
Regarding the `docsrs_bindings.rs` merge conflict probably best just to regenerate after rebasing.

src/types/array.rs
mszabo-wikia: That seems like a good solution, thanks.

kakserpom:
Reworked in https://github.com/extphprs/ext-php-rs/pull/631

Xenira:
Resolved in #631

Issue #335, opened 2024-10-23
Switch to Outer Macro pattern


Xenira:
## Motivation
Currently the macros depend on the line order of the code. Since recently the way the current macros work break rust analyzer. This was noted in #327.
A solution to this problem is to switch to the Outer Macro pattern, which allows to track the state of the macro without depending on the line order of the code.

## Design Detail

The main idea is to use a `#[php_module]` macro on a module that contains all the functions, classes, etc. that should be exported to PHP.
As everything in that module is the input to the macro, the macro can track the state of the module and generate the necessary code.

While parsing the 'pseudo' inner macros are evaluated similar to the current macros.

This would look something like this:
```rust
#[php_module]
pub mod module {
    #[php_function]
    fn func() { ... }

    #[php_class]
    pub struct Example {
        [...]
    }

    #[php_impl]
    impl Example {
        [...]
    }
}
```

## Drawbacks

The main drawback is that all functions, classes, etc. must be defined in the same module. This can make the code harder to read and maintain.
A workaround for this would be to include a `require!` macro that allows to include the code from another module. Functionality for this can be added using [macro_magic](https://crates.io/crates/macro_magic).
It should be noted, that `macro_magic` has this warning about that approach:
>Any types you may have brought into scope with use statements in the foreign module may or may not be available in their new context without additional use statements.

Another drawback is that this will break existing code that uses the old macro pattern.

## Rationale and Alternatives

The main alternative is to drop the `#[php_module]` macro and require the user to register all functions, classes, etc. manually (#329).
This gives the author more control over how the module is wired in the end. I guess this would be similar to how crates like `actix_web` do things.
While this would work, it would also be a breaking change and not that easy to migrate to as this approach requires some manual work by the author.
Using the Outer Macro pattern would in theory allow to support both patterns, but would make most cases easier.

Another alternative is to keep the current macro pattern and fix the issues with it. This would still depend on the line order of the code and has no guarantee that it will not break again in the future.

## Prior Art

[This youtube video](https://youtu.be/aEWbZxNCH0A?feature=shared&t=1989) explains the concept of Outer Macros.

Xenira:
@faassen

faassen:
Okay, I wasn't aware of this pattern, so I'm going to have to read up on it. 

[...]

PR #372, opened 2025-02-20
refactor(macro)!: change to builder pattern

Xenira:
BREAKING CHANGE: The old macros were dependent on execution order and have been causing trouble with language servers. They are replaced by a builder. See the migration guide at https://davidcole1340.github.io/ext-php-rs/migration-guides/v0.14.html for information on how to migrate.

Fixes: #99, #131, #327
Refs: #174, #335


Xenira:
Decided to go with builder pattern and move the outer macro back in priority. Would still like to have it, but there are some cases that really need a builder, so this is the foundation.

Xenira:
So this is a chonker...

Most of it is taken from #174 which already had some review. I tried to rebase onto current master, but it was to old to get that done cleanly.

I pulled what I could and integrated it with the new stuff and added stubs back in.

examples/hello_world.rs
```diff
@@ -0,0 +1,88 @@
+#![cfg_attr(windows, feature(abi_vectorcall))]
+use ext_php_rs::{constant::IntoConst, prelude::*, types::ZendClassObject};
+
+#[derive(Debug)]
+#[php_class]
+pub struct TestClass {
+    #[prop]
```
joelwurtz, line 7: i find this name not really consistent given other macros, maybe it would be nice to have a better one.

(Not necessarily in this PR, but before the next release at least, we can profit from this BC break to make things consistent ?)

examples/hello_world.rs
```diff
@@ -0,0 +1,88 @@
...
+    a: i32,
+    #[prop]
+    b: i32,
+}
+
+#[php_impl]
+impl TestClass {
+    #[rename("NEW_CONSTANT_NAME")]
```
joelwurtz, line 15: Same as `prop` macro, it's not really consistent and we could profit from the BC break to make this better IMO

joelwurtz (commented):
Make some comment about API and doc, like i have said in the original PR i think this is the best for the future. 

Did not look at the implementation, not that much time, but IMO if the API is good implementation can always be fixed later

examples/hello_world.rs
Xenira: So your suggestion would be something like `#[php_prop]`?

Not a 100% sure how I feel about that given that it isn't an actual macro and its simpler the way it is. On the other hand a user might not actually care about that distinction.

I'll open an issue for this.

Xenira: #391

Xenira: Maybe something like serde is doing would be a good fit.

`#[php(prop)]`
`#[php(rename = "foo")]`
[...]

joelwurtz: Yeah, i was thinking at something like this, not sur how doable it is since i'm not used to write derive macro

Xenira: Should be doable.

Ill resolve these conversations and add the suggestion to the issue.

joelwurtz (approved):

Xenira:
I'll just merge this now. Has been open long enough for comments.

Lets just wait a bit longer before the next release to find issues.

crates/macros/src/impl_.rs
```diff
@@ -88,172 +88,310 @@ pub struct AttrArgs {
...
+                    vec![#(#functions),*]
+                }
 
-pub fn parse_attribute(attr: &Attribute) -> Result<Option<ParsedAttribute>> {
-    let name = attr.path.to_token_stream().to_string();
-    let meta = attr
-        .parse_meta()
-        .map_err(|_| anyhow!("Unable to parse attribute."))?;
+                fn get_method_props<'a>(self) -> ::std::collections::HashMap<&'static str, ::ext_php_rs::props::Property<'a, #path>> {
+                    todo!()
```
joehoyle, line 362: It looks like this was never fixed, so actually getters / setters don't work at all right now

PR #396, opened 2025-03-09
chore(test): fix embed test on php 8.4

joelwurtz:
`zend_register_module_ex` change its signature in php 8.4

### DIFF
--- a/tests/module.rs
+++ b/tests/module.rs
@@ -10,7 +10,13 @@ use ext_php_rs::prelude::*;
 fn test_module() {
     Embed::run(|| {
         // Allow to load the module
-        unsafe { zend_register_module_ex(get_module()) };
+        cfg_if::cfg_if! {
+            if #[cfg(php84)] {
+                unsafe { zend_register_module_ex(get_module(), 2) };
+            } else {
+                unsafe { zend_register_module_ex(get_module()) };
+            }
+        }
 
         let result = Embed::eval("$foo = hello_world('foo');");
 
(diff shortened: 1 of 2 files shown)


Xenira:
Guess the pipeline should also run embedded for all supported php versions.

Not necessarily in this pr though.

Will have a closer look at the changes once I am back from work.

joelwurtz:
Yeah that's what i think so also, i try to look at the possibility of passing a build var to the dockerfile but did not see it (would avoid duplicating dockerfile)

tests/module.rs
```diff
@@ -10,7 +10,13 @@ use ext_php_rs::prelude::*;
 fn test_module() {
     Embed::run(|| {
         // Allow to load the module
-        unsafe { zend_register_module_ex(get_module()) };
+        cfg_if::cfg_if! {
+            if #[cfg(php84)] {
+                unsafe { zend_register_module_ex(get_module(), 2) };
```
Xenira, line 15:
```suggestion
            if #[cfg(php84)] {
                // Register as temporary (2) module
                unsafe { zend_register_module_ex(get_module(), 2) };
```

tests/module.rs
```diff
@@ -10,7 +10,13 @@ use ext_php_rs::prelude::*;
 fn test_module() {
     Embed::run(|| {
         // Allow to load the module
-        unsafe { zend_register_module_ex(get_module()) };
+        cfg_if::cfg_if! {
```
Xenira, line 13: I think we can import `cfg_if` here. looks cleaner than `cfg_if::cfg_if!`

Xenira (changes requested):
Looks good to me. Just 2 small nitpicks. Thank you!

Xenira (approved):

Issue #322, opened 2024-09-06
Docs/book link issue 


romanstingler:
There is a small annoyance in the docs.
The link for `Async futures` and `Async Impl` are the same and therefore in the book if you press the > (next page arrow) if you are on 'async impls' it jumps back to `5. Macros`

![image](https://github.com/user-attachments/assets/6375435f-787c-48f4-aece-492f7b8ed68a)

PR #399, opened 2025-03-11
create symlink for async_impl doc #322

yoramdelangen:


Xenira:
@yoramdelangen thanks for tying this. After looking at this more closely I think it is better to just move async into a different section, as it is an external crate `php-tokio`.

Done that in #402 and will close this.

yoramdelangen:
Way better solution!

Issue #391, opened 2025-02-28
Consider changing `#[prop]` and `#[rename]` 'macro' names


Xenira:
As the macro rework already is a big BC change we might want to consider using a more consistent naming scheme.

Maybe something like:
`#[php(prop)]`
`#[php(rename = "foo")]`
[...]

https://github.com/davidcole1340/ext-php-rs/pull/372#discussion_r1975739299

Xenira:
I don't like the idea of `#[php(hide)]` as that would default to public and things should be private by default.

Id rather use the rust visibility here, but would need to look at it more closely.

Xenira:
Did some testing on functions and something like this is easy and I kinda like it:

```rs
#[php_function)]
#[php(rename = "test_default", vis = "public")]
#[php(defaults(a = 5))]
pub fn test_default(a: i32) -> i32 {
    a
}
```

PR #400, opened 2025-03-12
chore(deps): update syn and darling

Xenira:
Update syn to 2.0.100
Update darling to 0.20


Xenira:
Guess this can be further improved by doing more `darling` parsing, but that is something for #391

PR #405, opened 2025-03-18
feat(sapi): add global module handling

joelwurtz:
Still a draft but i want to open this to discussion before going further.

I want to try what it would give to have a proper API for creating a SAPI module for PHP.

Current api is limited and was mainly for test.

The goal here would be to accept that there should only one SAPI at a time (even if we can shut it down properly and create it afterwards it does not really make sense as a feature, it's like having multiple main method)

In order to that i want to propose the following api : 

```rust
sapi::set_global_module(module: SapiModule) -> Result<(), SetSapiModuleError>;
sapi::get_global_module() -> Option<&'static SapiModule>;
```

`set_global_module` would set the global sapi_module variable and make the necessary step to initialize the module

I'm wondering if it should return a "guard" that would shutdown the module when dropped, or even lock installing a module ? (For me it's not necessary as long as we don't allow calling twice this function, otherwise it can be useful)

This method would throw an error if there is already an existing sapi module (at the moment i only check that the name is not null, not sure if there is a better way ?)

The `get_global_module` allow to retrieve the current installed sapi

Then someone could create a rust binary that act as a SAPI for PHP by doing something like : 

```rust
fn main() {
    let sapi_module = SapiBuilder::new("my-sapi", "My awesome sapi")
        ->build().expect("cannot build sapi module");
        
    ext_php_rs::embed::sapi::set_global_module(sapi_module).expect("cannot register sapi module");
    
    // we can now execute a php script safely
}
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM php:zts-bullseye
+
+# Install rust
+RUN apt-get update && apt-get install -y curl clang
+
+# Add app user with uid 1000
+RUN useradd -m -u 1000 app
+USER app
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+ENV PATH="/home/app/.cargo/bin:${PATH}"
--- a/allowed_bindings.rs
+++ b/allowed_bindings.rs
@@ -307,6 +307,7 @@ bind! {
     _zend_bailout,
     sapi_startup,
     sapi_shutdown,
+    sapi_module,
     php_module_startup,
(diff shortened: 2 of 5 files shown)


joelwurtz:
I will close this, did not see the sapi globals struct already existing i think it should use that instead, will think about a better api around it

Xenira:
@joelwurtz can you create an issue for this? Would like to include this in `0.14` if it is an API change and that way I can better track it.

joelwurtz:
Don't wait for this, not sure i will have the time soon to tackle this

Issue #397, opened 2025-03-10
Run embedded tests for all supported php versions


Xenira:
Guess the pipeline should also run embedded for all supported php versions.

_Originally posted by @Xenira in https://github.com/davidcole1340/ext-php-rs/issues/396#issuecomment-2709452918_

PR #407, opened 2025-03-21
chore(ci): simplify ci, test all features combination

joelwurtz:
THis should resolve https://github.com/davidcole1340/ext-php-rs/issues/397 and certainly other things

 * removed the specific embed / zts images
 * test on thread safe for macos / linux env
 
I added cargo hack (and an alias to config) to make the test in CI check that everything works given all possible combinations of features. Also runnable in local (if you have cargo hack) by doing : `cargo ci-test`


joelwurtz:
I'm wondering if it's necessary to test both in stable and nightly (lot of variations) ? Maybe only testing php8.4 + nts + nightly is enough ?

Xenira:
@joelwurtz unstable has pointed out some problems in the past and I would like to keep it for that reason.

Regarding the amount of variations, we should test all supported versions.
I'd like to reduce it as well, and things like 8.0 which is EOL could reduce it a little. I personally would like to drop it but till then I don't see much of a way to reduce it.

While CI takes a little time it is still reasonable.

joelwurtz:
Would having unstable only on one version (like done actually) would be enough ? Or should we do it on all versions ?

Also don't know why it does not work on windows, i can reproduce the problem on my windows system but it's only when clang > 15 (when using clang 15 it works) but looking at the script in CI it does not seems to install clang 15 actually so i don't know why it works ?

Xenira:
> Would having unstable only on one version (like done actually) would be enough ? Or should we do it on all versions ?

At least for windows we need all unstable, as stable is not supported afaik. On all others I would be fine with one unstable but full test on a nightly build.

> Also don't know why it does not work on windows, i can reproduce the problem on my windows system but it's only when clang > 15 (when using clang 15 it works) but looking at the script in CI it does not seems to install clang 15 actually so i don't know why it works ?

I don't have windows. So prob not much help there.

joelwurtz:
I'm closing this atm, too old, and didn't achieve what i want

PR #408, opened 2025-03-22
feat(zts): set lock per thread on zts build

joelwurtz:
Since on zts each thread gets its own globals (at the exception of sapi_module), i updated the lock to be on local thread (instead of static) so we can update / get globals without getting lock by another thread when in ZTS

### DIFF
--- a/src/zend/globals.rs
+++ b/src/zend/globals.rs
@@ -1,12 +1,12 @@
 //! Types related to the PHP executor, sapi and process globals.
 
+use parking_lot::{ArcRwLockReadGuard, ArcRwLockWriteGuard, RawRwLock, RwLock};
 use std::collections::HashMap;
 use std::ffi::CStr;
 use std::ops::{Deref, DerefMut};
 use std::slice;
 use std::str;
-
-use parking_lot::{const_rwlock, RwLock, RwLockReadGuard, RwLockWriteGuard};
+use std::sync::{Arc, LazyLock};
 
 use crate::boxed::ZBox;
 use crate::exception::PhpResult;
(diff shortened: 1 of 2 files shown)


src/zend/globals.rs
```diff
@@ -51,7 +51,11 @@ impl ExecutorGlobals {
         // return an invalid pointer.
         let globals = unsafe { ext_php_rs_executor_globals().as_ref() }
             .expect("Static executor globals were invalid");
-        let guard = GLOBALS_LOCK.read();
+        #[cfg(not(php_zts))]
+        let guard = lock::GLOBALS_LOCK.read_arc();
+        #[cfg(php_zts)]
+        let guard = lock::GLOBALS_LOCK.with(|l| l.read_arc());
```
Xenira, line 57: Use `cfg-if` for those. Makes the `if`/`else` nature of these more obvious.

Xenira (changes requested):
Thanks! Only have one nitpick :)

src/zend/globals.rs
joelwurtz: done

src/zend/globals.rs
```diff
@@ -67,7 +75,15 @@ impl ExecutorGlobals {
         // return an invalid pointer.
         let globals = unsafe { ext_php_rs_executor_globals().as_mut() }
             .expect("Static executor globals were invalid");
-        let guard = GLOBALS_LOCK.write();
+
+        cfg_if::cfg_if! {
+            if #[cfg(php82)] {
```
Xenira, line 80: Prob didn't mean `php82` here

joelwurtz: yeah bad cc sorry, should be good

Xenira (approved):

PR #412, opened 2025-04-01
fix(macro): Add missing separator pipe in flags

Norbytus:
While impl interface registration, find error in proc macros php_impl
Is missing separator for quote

### DIFF
--- a/crates/macros/src/impl_.rs
+++ b/crates/macros/src/impl_.rs
@@ -388,7 +388,7 @@ impl quote::ToTokens for FnBuilder {
             flags.push(quote! { ::ext_php_rs::flags::MethodFlags::Abstract });
         }
         quote! {
-            (#builder, #(#flags)*)
+            (#builder, #(#flags)|*)
         }
         .to_tokens(tokens);
     }


Xenira:
@Norbytus thanks for catching this.

I'll check if I can add a simple test to catch this going forward.

Norbytus:
> @Norbytus thanks for catching this.
> 
> 
> 
> I'll check if I can add a simple test to catch this going forward.

I actually try little bit rewrite method generation. Cause it's make some problem to add interface registration.

I make pr later or not)

Xenira:
> I actually try little bit rewrite method generation. Cause it's make some problem to add interface registration.

I just did a big change on how the macros work (#372). Some bugs are expected and there are some things that can be done better. Thats why 0.14 is not released yet.

If you plan a bigger change, please create an issue first though. That way possible solutions can be discussed and you do not waste your time on something that is going in the wrong direction.

> I make pr later or not)

Its usually better to create PRs as early as possible. That way you can get feedback earlier. Just mark it as draft to indicate that its not ready yet.

PR #419, opened 2025-04-06
fix: Messing static flags in php_impl proc macro

Norbytus:
If try class static method from php it throws exception
```
Uncaught Error: Non-static method Test::staticTest()
```

Cause in proc_macros missing usage of MethodFlags::Static

```rust
#[php_class]
#[derive(Default)]
struct Test {}

#[php_impl]
impl Test {
    #[php(constructor)]
    fn new() -> Self {
        Self {}
    }
    /// Declare as static method
    fn static_test() -> String {
        String::from("Hello")
    }
}
```

```php
<?php

$t = new Test;
Test::staticTest(); // Get exception
```

In PR i change FnBuilder from single additional modifier with flag abstract to Vec<Modifier>, Myabe we should make uniq Vec, idk)

### DIFF
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -210,6 +210,10 @@ impl TestClass {
     pub fn set_number(&mut self, number: i32) {
         self.number = number;
     }
+
+    pub fn static_call(name: String) -> String {
+        format!("Hello {name}")
+    }
 }
 
 #[php_function]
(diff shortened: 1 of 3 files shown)


Xenira:
Nice catch. Would you mind including an integration test as well.

Will have a closer look at the implementation this evening.

tests/src/lib.rs
```diff
@@ -210,6 +210,10 @@ impl TestClass {
     pub fn set_number(&mut self, number: i32) {
         self.number = number;
     }
+
+    pub fn static_call(name: String) -> String {
+        format!("Hello {}", name)
```
Xenira, line 215:
```suggestion
        format!("Hello {name}")
```

Xenira (changes requested):

Xenira (approved):

PR #420, opened 2025-04-06
refac: Make rename as Trait && separate to common rename, method_rename

Norbytus:
I see how you @Xenira changed renaming, maybe change it to trait for common rename things and method?

### DIFF
--- a/crates/macros/src/parsing.rs
+++ b/crates/macros/src/parsing.rs
@@ -1,5 +1,25 @@
 use darling::FromMeta;
 
+const MAGIC_METHOD: [&str; 17] = [
+    "__construct",
+    "__destruct",
+    "__call",
+    "__call_static",
+    "__get",
+    "__set",
+    "__isset",
+    "__unset",
+    "__sleep",
+    "__wakeup",
+    "__serialize",
+    "__unserialize",
+    "__to_string",
+    "__invoke",
+    "__set_state",
+    "__clone",
+    "__debug_info",
+];
+
 #[derive(Debug, FromMeta)]
 pub enum Visibility {
     #[darling(rename = "public")]
@@ -8,6 +28,14 @@ pub enum Visibility {
     Private,
     #[darling(rename = "protected")]
     Protected,
+}
+
+pub trait Rename {
+    fn rename(&self, rule: &RenameRule) -> String;
+}
+
+pub trait MethodRename: Rename {
+    fn rename_method(&self, rule: &RenameRule) -> String;
 }
 
 #[derive(FromMeta, Debug, Default)]
@@ -24,7 +52,7 @@ impl PhpRename {
                 let name = name.as_ref();
                 self.rename
                     .as_ref()
-                    .map_or_else(|| name.to_string(), |r| r.rename(name))
+                    .map_or_else(|| name.to_string(), |r| name.rename(r))
             },
             ToString::to_string,
         )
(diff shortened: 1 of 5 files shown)


crates/macros/src/parsing.rs
```diff
@@ -133,44 +167,26 @@ mod tests {
...
-        assert_eq!(pascal, RenameRule::Pascal.rename(original));
-        assert_eq!(snake, RenameRule::Snake.rename(original));
+        assert_eq!(original, original.rename(&RenameRule::None));
+        assert_eq!(camel, original.rename(&RenameRule::Camel));
+        assert_eq!(pascal, original.rename(&RenameRule::Pascal));
+        assert_eq!(snake, original.rename(&RenameRule::Snake));
         assert_eq!(
             screaming_snake,
-            RenameRule::ScreamingSnakeCase.rename(original)
+            original.rename(&RenameRule::ScreamingSnakeCase)
```
Xenira, line 189: Shouldn't this be `rename_method`. At least the test is called `test_rename_php_methods`

Xenira: Also probably want to add test for normal rename.

Xenira (changes requested):

crates/macros/src/parsing.rs
Norbytus: No problem.
One question how bad it will be, if i delete prefix test in tests?
test_rename_magic_method -> rename_magic_method

Xenira: Would be fine with me

Norbytus:
@Xenira Also i fix magic method name,
__to_string => __toString
and __call_static => __callStatic

https://www.php.net/manual/en/language.oop5.magic.php

I hope it not make any bc for old version

Xenira:
If I see this correctly this breaks the functionality. The idea was to rename `__to_string` to `__toString`

Norbytus:
> If I see this correctly this breaks the functionality. The idea was to rename `__to_string` to `__toString`

Ok, I will revert this changes

crates/macros/src/parsing.rs
Xenira: Sorry for the delay. Will have a look this evening.

Norbytus: No problem)

Xenira:
@Norbytus I added the explicit test cases back in, as I find those are usually better. Seems like they found a problem as well.

Do you want to have a look at it, or should I fix them.

Norbytus:
> @Norbytus I added the explicit test cases back in, as I find those are usually better. Seems like they found a problem as well.
> 
> Do you want to have a look at it, or should I fix them.

Ohhh, i will fix it. Thanks. I really miss there is mapping from __snake to __camel

crates/macros/src/parsing.rs
```diff
@@ -114,12 +114,23 @@ impl MethodRename for syn::Ident {
...
+        match rule {
+            RenameRule::None => self.to_string(),
+            _ => {
+                if MAGIC_METHOD.contains(self) {
+                    if self == &MAGIC_METHOD[12] {
+                        "__toString".to_string()
+                    } else if self == &MAGIC_METHOD[16] {
+                        "__debugInfo".to_string()
+                    } else if self == &MAGIC_METHOD[3] {
+                        "__callStatic".to_string()
```
Xenira, line 126: Don't like the index reference here. Changing the order of the items would change this as well without being apparent.

I think it would be better to either compare it against a static string or store the map in `MAGIC_METHOD`.

Xenira (changes requested):

crates/macros/src/parsing.rs
Norbytus: I agree, i wanna ask your advice about this

Xenira (approved):
Thanks for all your effort. Looks good to me :)

Norbytus:
> Thanks for all your effort. Looks good to me :)

Cool, i think im done. But i make two issue while write test)
I check theme later

Issue #422, opened 2025-04-07
Improve `#[php]` attribute documentation


usarise:
After commit https://github.com/davidcole1340/ext-php-rs/commit/90dfe330a775d1936eacd0b51c265a24be386121
php class name assignment broken.

Sample:
```rust
#[php_class(name = "Test\\Bug")]
pub struct TestBug {}

#[php_impl]
impl TestBug {
    const TEXT: &str = "issue";
}
```

In php script does not work `Test\Bug::TEXT`
and it works `TestBug::TEXT`

Xenira:
Did you try

```
#[php_class]
#[php(name = "Foo")]
```

https://github.com/davidcole1340/ext-php-rs/blob/master/guide/src/migration-guides/v0.14.md#classes

Xenira:
You are right. Documentation can be improved here

Xenira:
Good point. Will update that as well to be consistent.

PR #427, opened 2025-04-24
docs(macro): improve `name` vs `rename` documentation

Xenira:
Refs: #422


Xenira:
@usarise can you check if this is clear enough? Always open to suggestions.

Issue #189, opened 2022-11-23
#[prop] attribute should rename fields to camelCase by default


ju1ius:
Currently, the `#[prop]` attribute does not handle case conversion, so we must resort to:
```rust
#[php_class]
struct Foo {
  #[prop(rename = "isBar")]
  pub is_bar: bool,
}
```
which can quickly become tedious.

The `#[php_class]` attribute should accept a `rename_fields` argument similiar to `rename_methods` for `#[php_impl]`, and default to `camelCase`.

The `#[prop(rename)]` argument should be kept around for corner-cases like:
```rust
#[php_class]
struct Request {
  #[prop(rename = "isXMLHttpRequest")]
  pub is_xml_http_request: bool,
}
```
Although on might argue that it should be `#[prop(name = ..)]` and not `#[prop(rename = ..)]`...

For ergonomics, it could even be made into a global, project-wide, configuration option?

ju1ius:
Thinking about it, it should look more like :
```rust
#[php_class(field_case = "snake", method_case = "camel", constant_case = "upper")]
struct Foo;
```
This would ease the burden of verbose repeated configuration.

Dennis-Zhang-SH:
Well I think you could simply rename your function and use `#[allow(non_snake_case)]`
:)

PR #436, opened 2025-05-29
chore(macro)!: change rename defaults to match psr

Xenira:
BREAKING CHANGE: Methods and Properties are renamed to camelCase by default. Classes to PascalCase and constants to UPPER_CASE.

Refs: #189


joelwurtz:
It seems php functions are also renamed to camelCase by default where it should be snake_case no ?

Xenira:
@joelwurtz psr does not mention functions directly as far as I can tell. Seems both camel and snake case are possibilities here.

I like the idea of using snake here though.

Issue #262, opened 2023-08-04
Err: Attempted to access uninitialized class object


godruoyi:
When I want to custom the exception use the bellow code, that's copy from https://davidcole1340.github.io/ext-php-rs/macros/classes.html#example

```rust
#![cfg_attr(windows, feature(abi_vectorcall))]
extern crate ext_php_rs;

use ext_php_rs::prelude::*;
use ext_php_rs::{exception::PhpException, zend::ce};

#[php_class(name = "Redis\\Exception\\RedisException")]
#[extends(ce::exception())]
#[derive(Default)]
pub struct RedisException;

// Throw our newly created exception
#[php_function]
pub fn throw_exception() -> PhpResult<i32> {
    Err(PhpException::from_class::<RedisException>("Not good!".into()))
}

#[php_module]
pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
    module
}
```

tests/Unit/BasicTest.php
```
it('throw_exception', function () {
    throw_exception();
})->throws('Redis\Exception\RedisException', 'Not good!');
```

I get an error, 
```
$ cargo build
   Compiling myproject-php v0.1.0 (myproject)
    Finished dev [unoptimized + debuginfo] target(s) in 1.19s
    
$ php -d extension=../../target/debug/libmyproject_php.dylib vendor/bin/pest tests/Unit/BasicTest.php

thread '<unnamed>' panicked at 'Attempted to access uninitialized class object', ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ext-php-rs-0.10.1/src/types/class_object.rs:245:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
fish: Job 1, 'php -d extension=../../target/d…' terminated by signal SIGABRT (Abort)
```

Do you have any suggestions for me? I also want to create a PR to fix it, but I need some context or references first.

PR #313, opened 2024-03-07
fix: Allow to initialise objects

cv65kr:
Solves: https://github.com/davidcole1340/ext-php-rs/issues/262

PR #438, opened 2025-06-04
fix: fixed object access in object write/read/has_property handlers

alekitto:
Fixes: #262

Partially derived from #313, this commit fixes access to object properties from Rust code. This is needed to correctly initialize exception objects.


Xenira:
@alekitto Thanks for the PR. Was new to the project when #313 was created and forgot about it :/

Regarding the failed pipeline i have a pr open and will rebase this.

Xenira (approved):

PR #441, opened 2025-06-11
Fix SapiHeader::value when value contains a : character

Qard:
I forked ext-php-rs [here](https://github.com/platformatic/ext-php-rs) for use in [php-node](https://github.com/platformatic/php-node) as I needed to make a bunch of changes. I want to backport as much of that as possible now so _ideally_ we don't need to maintain a fork at all. Just starting small here with this small bug I fixed.

When a header value contains a : character, which _is_ in fact valid according to the spec, it will fail in the current code as it will only include the text between the : separating the header name and the first : within the value. This fixes that by only splitting into two pieces.

### DIFF
--- a/src/zend/globals.rs
+++ b/src/zend/globals.rs
@@ -532,7 +532,7 @@ impl<'a> SapiHeader {
     /// Returns the header value.
     #[must_use]
     pub fn value(&'a self) -> Option<&'a str> {
-        self.as_str().split(':').nth(1).map(str::trim)
+        self.as_str().split_once(':').map(|(_, value)| value.trim())
     }
 }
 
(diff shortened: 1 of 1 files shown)


Xenira:
@Qard first of all thank you for contributing your changes!

Could you add an integration test that covers this?

The lint pipeline is failing because the commit message does not adhere to conventional commits https://www.conventionalcommits.org/en/v1.0.0/

Qard:
I'm not yet sure exactly _how_ to write an integration test for it. Do you have some pointers? There seems to be very little test coverage for the globals interfaces presently, as far as I can tell. 🤔

Xenira:
Have a look at the existing tests https://github.com/davidcole1340/ext-php-rs/tree/master/tests/src/integration here.

If you have a specific question let me know.

I am currently working on increasing coverage in general, so there is still a lot missing. Integration tests don't show up in coverage report though.

EDIT: Misread that. Guess you wanted to know how to test the globals... Let me check

Qard:
Yeah, all I found for globals was https://github.com/davidcole1340/ext-php-rs/tree/master/tests/src/integration/globals which covers almost _none_ of the API surface, so I'm a bit unsure what a test should actually _look_ like, given there's not really any examples there.

Xenira:
@Qard well, looking at it seems like we would need to expand the sapi tests by a decent amount.

At this point a simple unit test should cover this. I added that. Could you check if this covers your case?

src/zend/globals.rs
```diff
@@ -532,7 +532,7 @@ impl<'a> SapiHeader {
     /// Returns the header value.
     #[must_use]
     pub fn value(&'a self) -> Option<&'a str> {
-        self.as_str().split(':').nth(1).map(str::trim)
+        self.as_str().splitn(2, ':').nth(1).map(str::trim)
```
Xenira:
```suggestion
        self.as_str().split_once(':').map(|(_, value)| value.trim())
```

Xenira (changes requested):

Qard:
Yep, that covers it. I'll see about fixing the commit messages for this now. Thanks for the help!

I'll have some more PRs coming over the next couple days. I want to get as much as possible copied over from [these commits](https://github.com/platformatic/ext-php-rs/compare/828338d5db8fa72c12b1c7bb2498a5410836e2dc...HEAD) so I don't need to be running a fork, and so others can benefit from the changes I've made. Much of what's in there will likely need a bit more polish, or _at least_ proper tests and docs, so I'll get through them all one at a time. 🙂

Xenira:
Much appreciated. Sapi needs some love :)

You can always just delegate when you don't have time to polish some parts. Not that much time either, but always happy to help.

Xenira (approved):

PR #442, opened 2025-06-11
Add IniBuilder

Qard:
This might only be useful in combination with the `embed` feature so it may make sense to gate this behind that feature, but it doesn't _depend_ on it in any way--it _can_ exist without it. I left the feature-gating out, but I could change that if you prefer.


allowed_bindings.rs
Xenira, line 1: Updated `allowed_bindings.rs` without updating `docs.rs`.

See https://github.com/davidcole1340/ext-php-rs/blob/master/allowed_bindings.rs#L8-L20 for details.

src/builders/mod.rs
```diff
@@ -3,12 +3,14 @@
...
 mod class;
 mod function;
+mod ini;
 mod module;
 #[cfg(feature = "embed")]
 mod sapi;
 
 pub use class::ClassBuilder;
 pub use function::FunctionBuilder;
+pub use ini::{IniBuilder};
```
Xenira:
```suggestion
pub use ini::IniBuilder;
```

Xenira (changes requested):
Haven't had a detailed look yet. But looks good.

I think we can keep it without a feature guard for now.

Xenira:
Would you mind adding documentation in the `/guide`?

And thanks again for contributing!

Qard:
Yep, I can add a guide for that. Where do you think it should go?

docsrs_bindings.rs
Xenira, line 1: Ahhh god. I really need to find a better way of handling this...

No one knows how to properly generate this file. I have a pretty good guess by now. Ill try to regenerate that for you. Sorry about that.

Will improve the docs while I am at it. Not sure if today though.

Qard: The path listed to copy didn't exist, so I had to go find the file, but I guess it's not the right one? Am I understanding right that there's two bindings files generated by the build.rs, one for the actual ffi use and a separate one for the docs there? I guess I grabbed the ffi one?

Xenira: Don't worry about it for now. Gonna need a reproducible way to generate this.

Depending on how your version of PHP is build the file contents change.

Just mocked up a docker build that can create this file automatically and check it in our ci.

Qard: Ah, cool. Thanks for looking into that! 🙂

Xenira: @Qard just merged the tooling to regenerate.

Can you rebase and check if its working for you?

```bash
tools/update_bindings.sh
```

Qard: Done! Appears to be working, though it edited a couple extra lines which I'm not sure if it was supposed to. Possibly due to the architecture difference? (I'm on M1 Mac, which is aarch64.)

Qard:
Seems the ini builder API didn't exist until 8.2. I've added compile gates to `builders/mod.rs`, but I'm not sure how to do the equivalent for the header include in the `wrapper.h` file. 🤔

Xenira:
> Seems the ini builder API didn't exist until 8.2. I've added compile gates to `builders/mod.rs`, but I'm not sure how to do the equivalent for the header include in the `wrapper.h` file. 🤔

Guess you could pass that in here: https://github.com/davidcole1340/ext-php-rs/blob/cafaecc692c474281925716f256f0842e4580b28/build.rs#L177

See windows build: https://github.com/davidcole1340/ext-php-rs/blob/cafaecc692c474281925716f256f0842e4580b28/windows_build.rs#L53

Qard:
I did some restructuring of the build file to include an ApiVersion enum which can be used for managing both the cfg flag names _and_ C define names in the stacked form it was using before, so we can do `EXT_PHP_RS_PHP_82` to gate C which depends on 8.2 minimum. 🙂

Qard:
This keeps failing with missing symbols for some reason. Not sure why yet. 🤔

Xenira:
I can have a look if you like. Probably later this week though.

Qard:
If you find some free time before I get to it, feel free. I'll be a bit busy this week too, so not sure when I'll get to it. 😅

Qard:
The issue seems to have been that php_ini_builder.h actually depends on SAPI things internally, so it needs to be gated on the `embed` feature actually. That's fine, since it's mainly embed users that'd want to use it anyway. 🤷🏻

Qard:
Hmm...not sure what that docsrs conflict is about--I don't get that change if I try to generate it locally. 🤔

ptondereau:
> Hmm...not sure what that docsrs conflict is about--I don't get that change if I try to generate it locally. 🤔 

Did you check if a new patch version of PHP 8.4 landed or maybe some commit has been added to 8.4 branch?

Qard:
If a new patch landed though, wouldn't the tooling get it whether it's running locally or in CI? 🤔

Qard:
Yeah, appears the tooling pulled a new patch in CI but local docker was caching the build layer for that.

Now clippy is yelling at me about a bunch of things that didn't change in this PR...did it pull an update that changed its rules or something? 🤔

Qard:
Ah, Rust 1.88 dropped a few days ago, and with it a new clippy update. We've got some new lints to sort out. 😅 

Looks like #460 already deals with that, so I'll just rebase this when that lands.

ptondereau:
> Ah, Rust 1.88 dropped a few days ago, and with it a new clippy update. We've got some new lints to sort out. 😅
> 
> Looks like #460 already deals with that, so I'll just rebase this when that lands.

I was about to write the same statement, haha 😅 
Yes, you felt into the update window of PHP and Rust

ptondereau:
@Qard merged the clippy PR, you can rebase

Qard:
Thanks! Rebased now. 🙂

Qard:
I still seem to be getting yelled at by clippy for unrelated things. 🤔 

The issues appear to be fixed by the clippy allows in #456, so maybe I'll just wait for that to land too? 😅

build.rs
```diff
@@ -249,19 +260,90 @@ fn generate_bindings(defines: &[(&str, &str)], includes: &[PathBuf]) -> Result<S
...
+
+impl TryFrom<u32> for ApiVersion {
+    type Error = Error;
+
+    fn try_from(version: u32) -> Result<Self, Self::Error> {
+        match version {
+            x if (2020_09_30..2021_09_02).contains(&x) => Ok(ApiVersion::Php80),
+            x if (2021_09_02..2022_08_29).contains(&x) => Ok(ApiVersion::Php81),
+            x if (2022_08_29..2023_08_31).contains(&x) => Ok(ApiVersion::Php82),
+            x if (2023_08_31..2024_09_24).contains(&x) => Ok(ApiVersion::Php83),
```
Xenira: Not as pretty, but this way we don't need to define magic numbers multiple times

```rs
match version {
    x if ((ApiVersion::Php80 as u32)..(ApiVersion::Php81 as u32)).contains(&x) => Ok(ApiVersion::Php80),
    [...]
}
```

src/builders/ini.rs
```diff
@@ -0,0 +1,278 @@
...
+        &self.0
+    }
+}
+
+impl Drop for CStringScope {
+    fn drop(&mut self) {
+        // Convert back to a CString to ensure it gets dropped
+        drop(unsafe { CString::from_raw(self.0) });
+    }
+}
```
Xenira: Do you want to extract this into its own file to make it reusable?

tests/src/integration/mod.rs
```diff
@@ -36,6 +37,43 @@ mod test {
         });
     }
 
+    /// Finds the location of an executable `name`.
+    #[must_use]
+    pub fn find_executable(name: &str) -> Option<PathBuf> {
+        const WHICH: &str = if cfg!(windows) { "where" } else { "which" };
+        let cmd = Command::new(WHICH).arg(name).output().ok()?;
+        if cmd.status.success() {
+            let stdout = String::from_utf8_lossy(&cmd.stdout);
```
Xenira: shouldn't we fail here if the result contains invalid characters and not just drop them silently?

Xenira:
Also needs a rebase and regeneration of the `allowed_bindings.rs` due to conflicts.

src/builders/ini.rs
```diff
@@ -0,0 +1,278 @@
...
+        assert!(!ini.is_null());
+        assert_eq!(
+            unsafe {
+                let data = slice::from_raw_parts(ini.cast::<u8>(), builder.length);
+                str::from_utf8(data)
+            },
+            Ok("grault=garply\n")
+        );
+    }
+}
```
ptondereau, line 293: maybe adding more tests to make sure:

```rust
#[test]
fn test_null_byte_error() {
    let mut builder = IniBuilder::new();
    assert!(builder.prepend("key=val\0ue").is_err());
}

#[test]
fn test_empty_values() {
    let mut builder = IniBuilder::new();
    assert!(builder.unquoted("", "").is_ok());
}
```

ptondereau (commented):
I think we also miss the opportunity to add an integration test by supporting setting `ini_entries` in [`SapiBuilder`](https://github.com/davidcole1340/ext-php-rs/blob/master/src/builders/sapi.rs).

in https://github.com/davidcole1340/ext-php-rs/blob/060c8bfb2dacffe35cffb558d4f3738d86a5d34c/tests/sapi.rs

We could have in addition something like this:

```rust
#[test]
fn test_sapi() {
    let mut ini = IniBuilder::new();
    ini.define("foo=bar").unwrap();
    ini.quoted("memory_limit", "128M").unwrap();

    let mut builder = SapiBuilder::new("test", "Test");
    builder = builder.ini_entries(ini.finish());
// ...    
```

src/builders/ini.rs
ptondereau: +1, it's very useful

Qard: Yep, definitely could do that. Where do you think it should go? 🤔

tests/src/integration/mod.rs
Qard: Possibly? I just copy/pasted this out of build.rs to get the PHP env var working here too as I needed it for my local dev. It's only used in `find_php` anyway, which returns a `Result<...>` itself, so trivially converted. 🙂

ptondereau (approved):

Qard:
I've updated to include all the feedback except moving the `CStringScope`. Just need to figure out where to put that. 😅

src/builders/sapi.rs
```diff
@@ -709,6 +727,25 @@ mod test {
...
+            .ini_entries(ini)
+            .build()
+            .expect("should build sapi module");
+
+        assert!(!sapi.ini_entries.is_null());
+        assert_eq!(
+            unsafe { CStr::from_ptr(sapi.ini_entries) },
+            c"foo=bar\nmemory_limit=\"128M\"\n"
+        );
+    }
```
ptondereau, line 748: as it require `embed` I don't know if this test can be executed/pass without the feature. Maybe skipping if `not(embed)`

Qard: The `SapiBuilder` _itself_ requires the embed feature, so this module wouldn't even be _included_ without it. `IniBuilder` _does_ require php82+ though, so I've gated it on that now. 🙂

Qard:
I moved that to `src/util/cstring_scope.rs` now. Anything else that needs doing, or is this good to land now? 🙂

PR #445, opened 2025-06-12
feat(globals): add CompilerGlobals

Qard:
I've exposed the CompilerGlobals interface. It's sometimes useful to access when building a SAPI, so it would be helpful for that to be exposed. 🙂


Xenira (approved):

ptondereau:
> I've exposed the CompilerGlobals interface. It's sometimes useful to access when building a SAPI, so it would be helpful for that to be expose. 🙂 

By building a SAPI, what do you mean? I'm bit curious 😅

Qard:
I used this to build [php-node](https://github.com/platformatic/php-node) which embeds PHP as a request handler inside Node.js in a worker thread pool. So not an extension, a server using the SAPI tools. I had to add some extra pieces to make that work and am in the process of backporting those now. 🙂

Issue #443, opened 2025-06-11
Automate `docsrs_bindings.rs` generation


Xenira:
Ahhh god. I really need to find a better way of handling this...

No one knows how to properly generate this file. I have a pretty good guess by now. Ill try to regenerate that for you. Sorry about that.

Will improve the docs while I am at it. Not sure if today though.

_Originally posted by @Xenira in https://github.com/davidcole1340/ext-php-rs/pull/442#discussion_r2140878734_

PR #446, opened 2025-06-12
chore(bindings): add tooling to generate `docsrs_bindings.rs`

Xenira:
Refs: #443

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,39 @@
+FROM rust:latest AS base
+ARG PHP_VERSION=8.4
+WORKDIR /tmp
+RUN <<EOF
+set -e
+
+apt update -y
+apt install -y \
+  libclang-dev \
+  bison \
+  re2c
+
+# Build PHP
+git clone --depth 1 -b PHP-${PHP_VERSION} https://github.com/php/php-src.git
+cd php-src
+# by default you will be on the master branch, which is the current
+# development version. You can check out a stable branch instead:
+./buildconf
+./configure \
+    --enable-debug \
+    --disable-all --disable-cgi
+make -j "$(nproc)"
+make install
+EOF
+
+FROM base AS docsrs_bindings_builder
+WORKDIR /src
+RUN rustup component add rustfmt
+RUN --mount=type=bind,target=/src,rw <<EOF
+set -e
+cargo clean
+cargo build
+cp target/debug/build/ext-php-rs-*/out/bindings.rs /docsrs_bindings.rs
+rustfmt /docsrs_bindings.rs
+EOF
+ENTRYPOINT ["/generate.sh"]
+
+FROM scratch AS docsrs_bindings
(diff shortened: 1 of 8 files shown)


tooling/update_bindings.sh
```diff
@@ -0,0 +1,12 @@
...
+# This script updates the bindings for the docs.rs documentation.
+# It should be run from the root of the repository.
+#
+# This script requires docker to be installed and running.
+set -e
+
+docker buildx build \
+  --target docsrs_bindings \
+  -o type=local,dest=. \
+  --build-arg PHP_VERSION=8.3 \
```
ptondereau: Maybe check against the last php version like 8.4?

Xenira: Sure, we could update it with this change

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# development version. You can check out a stable branch instead:
+./buildconf
+./configure \
+    --enable-debug \
+    --disable-all --disable-cgi
+make -j "$(nproc)"
+make install
+EOF
+
+FROM base AS docsrs_bindings_bilder
```
ptondereau:
```suggestion
FROM base AS docsrs_bindings_builder
```

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+set -e
+cargo clean
+cargo build
+cp target/debug/build/ext-php-rs-*/out/bindings.rs /docsrs_bindings.rs
+rustfmt /docsrs_bindings.rs
+EOF
+ENTRYPOINT ["/generate.sh"]
+
+FROM scratch AS docsrs_bindings
+COPY --from=docsrs_bindings_bilder /docsrs_bindings.rs /
```
ptondereau:
```suggestion
COPY --from=docsrs_bindings_builder /docsrs_bindings.rs /
```

Xenira:
> Great work BTW! It was a struggle point in the past 😅

Thanks. Yeah came up a lot :face_with_diagonal_mouth:

tooling/update_bindings.sh
Xenira: Will do the update separately, as that would require some more pipeline changes

Qard:
I think this might have an architecture problem. I updated #442, #444 and #445 with it and all had a few slight changes which appear to be unrelated to the PR changes and look an awful lot like architectural differences--differently sized types, mainly.

Xenira:
> I think this might have an architecture problem. I updated #442, #444 and #445 with it and all had a few slight changes which appear to be unrelated to the PR changes and look an awful lot like architectural differences--differently sized types, mainly.

@Qard Yes, you are right. Can you try adding a `--platform linux/amd64` to the docker build command in `tools/update_bindings.sh`. In my testing I got the same diff as you when building for arm.

Qard:
Yep, that appears to do it. I pushed an update to #444 which includes that change and the bindings look correct now. 🙂

PR #450, opened 2025-06-14
test: improve test reliability and ease of use

Qard:
This disables automatic filesystem-based test inclusion, opting for manual inclusion so the guides, which rely on additional features, can skip running unless all their required features are enabled.

This also wraps a few tests with `Embed::run(...)` as they may crash if the PHP environment is not fully set up. I think that is specific to ZTS builds as some things are not set up until the PHP bootstrap functions get called.


Xenira (approved):
Have to test this tomorrow, but looks good to me.

Thank you.

Issue #454, opened 2025-06-17
Allow negative array indices


Xenira:

kakserpom:
zend_hash_index* functions that are currently in use:
```
pub fn zend_hash_index_update(ht: *mut HashTable, h: zend_ulong, pData: *mut zval)
        -> *mut zval;
pub fn zend_hash_index_update(ht: *mut HashTable, h: zend_ulong, pData: *mut zval)
        -> *mut zval;
pub fn zend_hash_index_find(ht: *const HashTable, h: zend_ulong) -> *mut zval;
```

kakserpom:
So it looks more like a Zend API bug. Is there a workaround?

kakserpom:
I've opened an issue — https://github.com/php/php-src/issues/18872

kakserpom:
PR — https://github.com/davidcole1340/ext-php-rs/pull/456

Issue #453, opened 2025-06-17
Funny business with zend_array


kakserpom:
```rust
#[php_function]
pub fn make_a_peculiar_array() -> Result<Zval> {
    let mut array = zend_array::with_capacity(5);
    array.insert_at_index(1, false)?;
    array.insert("1", true)?;
    array.into_zval(false)
}
```

```php
<?php
$array = make_a_peculiar_array();
var_dump([
    "'1' exists" => array_key_exists("1", $array),
    "'1' value" => $array["1"],
    'count($array)' => count($array)
]);
unset($array["1"]);
var_dump([
    "array_key_exists('1', \$array)" => array_key_exists('1', $array),
    "isset(\$array['1'])" => isset($array['1']),
    'count($array)' => count($array),
    '$array' => $array,
]);
```

Output:

```
array(3) {
  ["'1' exists"]=>
  bool(true)
  ["'1' value"]=>
  bool(false)
  ["count($array)"]=>
  int(2)
}
array(4) {
  ["array_key_exists('1', $array)"]=>
  bool(false)
  ["isset($array['1'])"]=>
  bool(false)
  ["count($array)"]=>
  int(1)
  ["$array"]=>
  array(1) {
    ["1"]=>
    bool(true)
  }
}
```

Xenira:
> — negative array indexes are a thing in PHP.

I hate PHP with a passion for their "arrays". Wish they had arrays... Actually what got me started contributing here.

I will have a look, but prob. not this release (unless someone contributes), as I don't think this is a regression.

kakserpom:
Yeah, man, so do I. There should be dictionaries and vectors instead.

It's a pretty straightforward fix, I am working on it.

kakserpom:
I spent more time trying to figure out why 'master' is not working. It compiles okay, but no functions are getting exported.

PR #456, opened 2025-06-17
fix(array): cast numeric string keys into `zend_ulong` and allow negative keys

kakserpom:
https://github.com/davidcole1340/ext-php-rs/issues/453

### DIFF
--- a/tests/src/integration/array/mod.rs
+++ b/tests/src/integration/array/mod.rs
@@ -1,6 +1,9 @@
 use std::collections::HashMap;
 
-use ext_php_rs::{php_function, prelude::ModuleBuilder, wrap_function};
+use ext_php_rs::{
+    convert::IntoZval, ffi::HashTable, php_function, prelude::ModuleBuilder, types::Zval,
+    wrap_function,
+};
 
 #[php_function]
 pub fn test_array(a: Vec<String>) -> Vec<String> {
@@ -12,10 +15,23 @@ pub fn test_array_assoc(a: HashMap<String, String>) -> HashMap<String, String> {
     a
 }
 
+#[php_function]
+pub fn test_array_keys() -> Zval {
+    let mut ht = HashTable::new();
+    ht.insert(-42, "foo").unwrap();
+    ht.insert(0, "bar").unwrap();
+    ht.insert(5, "baz").unwrap();
+    ht.insert("10", "qux").unwrap();
+    ht.insert("quux", "quuux").unwrap();
+
+    ht.into_zval(false).unwrap()
+}
+
 pub fn build_module(builder: ModuleBuilder) -> ModuleBuilder {
     builder
         .function(wrap_function!(test_array))
         .function(wrap_function!(test_array_assoc))
+        .function(wrap_function!(test_array_keys))
 }
 
 #[cfg(test)]


src/types/array.rs
```diff
@@ -378,7 +379,11 @@ impl ZendHashTable {
         V: IntoZval,
```
Xenira, line 505: Maybe add a generic `Into<ArrayKey>` here and work with that

Xenira (commented):
Didn't have a detailed look yet, but remembered that we have an `ArrayKey` enum. Might want to utilise it here.

src/types/array.rs
kakserpom: ~~It would be great, but we can't use it  (`ArrayKey::Long(i64)`) until we fix https://github.com/davidcole1340/ext-php-rs/issues/454.~~

kakserpom: I've added ArrayKeyRef.

kakserpom:
PR solves both https://github.com/davidcole1340/ext-php-rs/issues/453 and https://github.com/davidcole1340/ext-php-rs/issues/454.

Xenira:
@kakserpom Would you mind adding integration and unit tests?

Xenira:
@kakserpom just a hint for commit messages you can use [convco](https://convco.github.io)

kakserpom:
@Xenira Ordnung über alles? 😁

Xenira:
@kakserpom keep it in english pls 😉

kakserpom:
Squash the commits, please. @davidcole1340 @Xenira

kakserpom:
@Qard done

Qard (commented):
What about all the #[allow(clippy::mut_from_ref)] lines? Are those actually necessary?

kakserpom:
@Qard yes, otherwise Clippy shows warnings because of &self instead of &mut self in function signatures.

Qard:
Ah, so it's intentional to leave it then? Just wanted to be sure we weren't just forgetting to un-silence clippy in some places it might have legitimate warnings. 😅

kakserpom:
Yes, It's intentional. I've enabled clippy::pedantic in another PR and quite a few warnings appeared, some of them led to actual fixes and some had to be silenced.

Qard (approved):
The clippy changes I guess aren't related to this PR, but necessary to get CI to pass, so that's fine.

Everything LGTM. Hopefully this can land soon to unblock other things. 🙂

Xenira:
@kakserpom @Qard Regarding the clippy issues: Our usual procedure is to create a mr fixing those issues and rebase affected MRs. I have created one for the current issues, except the ones in #461, as those require some more investigation and possible breaking changes.

@kakserpom could you remove those ignores and rebase once #474 is merged?

src/types/array.rs
```diff
@@ -25,6 +17,15 @@ use crate::{
...
 };
+use std::str::FromStr;
+use std::{
+    collections::HashMap,
+    convert::{TryFrom, TryInto},
+    ffi::CString,
+    fmt::{Debug, Display},
+    iter::FromIterator,
+    ptr,
+};
```
Xenira: These can be merged

src/types/array.rs
```diff
@@ -1,15 +1,7 @@
...
-use std::{
-    collections::HashMap,
-    convert::{TryFrom, TryInto},
-    ffi::CString,
-    fmt::{Debug, Display},
-    iter::FromIterator,
-    ptr,
-};
-
+use crate::ffi::zend_ulong;
```
Xenira: This can be merged below.

We usually have the crate use statement below other ones. Maybe change this back.

src/types/array.rs
```diff
@@ -606,15 +774,23 @@ pub struct Iter<'a> {
 }
 
 /// Represents the key of a PHP array, which can be either a long or a string.
-#[derive(Debug, PartialEq)]
-pub enum ArrayKey {
+#[derive(Debug, Clone, PartialEq)]
+pub enum ArrayKey<'a> {
```
Xenira, line 783: Add a comment here to explain the unsigned cast. As the cast happens at multiple places it would be overkill to have it at all those places. Guess this is the best place for it. If you have a better place feel free to place it there.

crates/cli/src/lib.rs
```diff
@@ -92,6 +92,7 @@ enum Args {
 
 #[allow(clippy::struct_excessive_bools)]
 #[derive(Parser)]
+#[allow(clippy::struct_excessive_bools)]
```
Xenira:
```suggestion
```

src/types/array.rs
```diff
@@ -277,8 +356,12 @@ impl ZendHashTable {
     // hashtable while only having a reference to it. #461
     #[allow(clippy::mut_from_ref)]
     #[must_use]
-    pub fn get_index_mut(&self, key: u64) -> Option<&mut Zval> {
-        unsafe { zend_hash_index_find(self, key).as_mut() }
+    #[allow(clippy::mut_from_ref)]
```
Xenira:
```suggestion
```

src/types/array.rs
```diff
@@ -219,9 +257,47 @@ impl ZendHashTable {
     // hashtable while only having a reference to it. #461
     #[allow(clippy::mut_from_ref)]
     #[must_use]
-    pub fn get_mut(&self, key: &'_ str) -> Option<&mut Zval> {
-        let str = CString::new(key).ok()?;
-        unsafe { zend_hash_str_find(self, str.as_ptr(), key.len() as _).as_mut() }
+    #[allow(clippy::mut_from_ref)]
```
Xenira:
```suggestion
```

src/types/array.rs
kakserpom: `cargo fmt` keeps doing this

Xenira: interesting... let me check

kakserpom:
@Xenira
> Could you also add integration tests for negative and 'string number' indices?

I'd rather someone else do it because I am on a trip until the end of the week. It's pretty trivial.

src/types/array.rs
Xenira: Ahhh, you were missing the empty line between them. Fixed it.

Xenira:
> I'd rather someone do it because I am on a trip until the end of the week. It's pretty trivial.

Sure, will add them :)

Xenira (approved):

PR #460, opened 2025-06-20
chore(clippy): flowing lifetimes warning and clippy::mut_from_ref

kakserpom:
- lifetime flowing from input to output with different syntax can be confusing
 - mutable borrow from immutable input(s)
 - chore(clippy): warning: implicit borrow as raw pointer

### DIFF
--- a/src/builders/class.rs
+++ b/src/builders/class.rs
@@ -303,7 +303,7 @@ impl ClassBuilder {
 
         let class = unsafe {
             zend_register_internal_class_ex(
-                &mut self.ce,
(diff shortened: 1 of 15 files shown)


Xenira:
@kakserpom `cargo fmt --all` and the pipeline should be green :)

kakserpom:
@Xenira, it cannot fix `mutable borrow from immutable input(s)` automagically. So we need to change it or suppress it.

Qard (approved):

ptondereau (approved):

Issue #452, opened 2025-06-17
cargo php --features


kakserpom:
This would be useful. Example:

`cargo php install --features postgres mysql --release --yes`

PR #463, opened 2025-06-20
feat(cargo-php): --features, --all-features, --no-default-features

kakserpom:
https://github.com/davidcole1340/ext-php-rs/issues/452

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -111,6 +111,12 @@ struct Install {
     /// the directory the command is called.
     #[arg(long)]
     manifest: Option<PathBuf>,
+    #[arg(short = 'F', long, num_args = 1..)]
+    features: Option<Vec<String>>,
+    #[arg(long)]
+    all_features: bool,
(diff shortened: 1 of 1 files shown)


crates/cli/src/lib.rs
```diff
@@ -469,17 +481,38 @@ fn find_ext(manifest: Option<&PathBuf>) -> AResult<cargo_metadata::Target> {
...
+        }
+    }
+
+    if all_features {
+       cmd.arg("--all-features");
+    }
+
+    if no_default_features {
+        cmd.arg("--no-default-features");
+    }
```
ptondereau: I don't know if it is an edge case, but this kind like of exclusives flags when these two cannot co-exist at the same time

kakserpom: Well I am letting cargo handle that. Just passing those as-is.

ptondereau:
@kakserpom can you fix `fmt` issues please before merging?

kakserpom:
@ptondereau  done

Issue #467, opened 2025-06-20
Doc comment bug in #[php_class]


kakserpom:
The error only occurs if there's more than one line in a doc comment.

I spent more than an hour to figure out what caused it.

```
use ext_php_rs::prelude::*;
/// A
/// B
#[php_class]
pub struct Foo {
}
#[php_module]
pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
    module.class::<Foo>()
}
```

```
error: expected one of `,`, `.`, `;`, `?`, `]`, or an operator, found `" B"`
 --> src/lib.rs:4:1
  |
4 | #[php_class]
  | ^^^^^^^^^^^^
  | |
  | expected one of `,`, `.`, `;`, `?`, `]`, or an operator
  | while parsing this item list starting here
  | the item list ends here
  |
  = note: this error originates in the attribute macro `php_class` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0046]: not all trait items implemented, missing: `get_metadata`, `get_properties`, `method_builders`, `constructor`, `constants`
 --> src/lib.rs:4:1
  |
4 | #[php_class]
  | ^^^^^^^^^^^^ missing `get_metadata`, `get_properties`, `method_builders`, `constructor`, `constants` in implementation
  |
  = note: this error originates in the attribute macro `php_class` (in Nightly builds, run with -Z macro-backtrace for more info)
  = help: implement the missing item: `fn get_metadata() -> &'static ClassMetadata<Self> { todo!() }`
  = help: implement the missing item: `fn get_properties() -> HashMap<&'static str, PropertyInfo<'a, Self>> { todo!() }`
  = help: implement the missing item: `fn method_builders() -> std::vec::Vec<(FunctionBuilder<'static>, MethodFlags)> { todo!() }`
  = help: implement the missing item: `fn constructor() -> std::option::Option<ConstructorMeta<Self>> { todo!() }`
  = help: implement the missing item: `fn constants() -> &'static [(&'static str, &'static (dyn IntoZvalDyn + 'static), &'static [&'static str])] { todo!() }`

For more information about this error, try `rustc --explain E0046`.
error: could not compile `untitled5` (lib) due to 2 previous errors
Error: Compilation failed, cancelling installation.
```

Xenira:
Ohhh god. Thanks for catching this. What a little comma can do...

PR #468, opened 2025-06-20
fix(macro): fix syntax error for `#[php_class]` with multiline doc comments

Xenira:
Refs: #467


Xenira:
@kakserpom can you check if this fixed it for you?

kakserpom:
@Xenira yes, it did.

PR #471, opened 2025-06-26
feat: expand SapiBuilder

Qard:
I've expanded the SapiBuilder to cover _most_ of the API surface area. This also includes tests to verify the struct fields are set properly.

Do we want some functionality tests too? That'll be a bit more work as it's a bit complicated to set up a fully-functional environment properly.

I also adjusted some more tests that needed the `Embed::run(...)` wrapper to be ZTS-safe. I can pull those out to a separate PR though, if you prefer.


Qard:
@Xenira This is the main SAPI-related change I need to be able to drop my fork. How does this look? Anything more you'd want from it?

Xenira:
@Qard Thank you. Was on vacation the past few days and need to catch up as a lot happened while I was away. Will have a look as soon as possible.

Qard:
No worries. I hope it was a nice vacation! 🙂

src/builders/sapi.rs
```diff
@@ -1,19 +1,54 @@
...
+};
+use crate::types::Zval;
 use crate::{embed::SapiModule, error::Result};
 
-use std::ffi::{c_char, c_void};
+use std::ffi::{c_char, c_int, c_void};
 use std::{ffi::CString, ptr};
 
 /// Builder for `SapiModule`s
+///
```
Xenira, line 13:
```suggestion
///
/// # Example
///
```

PR #472, opened 2025-06-26
chore: enable zts Linux and macOS builds in CI

Qard:
The comment seems to not be true anymore, setup-php has supported it for a couple years now.


ptondereau (approved):
thank you

Issue #447, opened 2025-06-12
Update docsrs bindings to PHP 8.4


Xenira:
Maybe check against the last php version like 8.4?

_Originally posted by @ptondereau in https://github.com/davidcole1340/ext-php-rs/pull/446#discussion_r2143322187_

ptondereau:
I think it's already generated from 8.4

Xenira:
Nope, its 8.3

PR #478, opened 2025-07-02
chore(bindings): update `docsrs_bindings.rs` to PHP 8.4

Xenira:
Refs: #447


ptondereau:
Nice one and maybe upgrading bindgen?

Xenira:
Bindgen is making trouble upgrading. Will investigate after next release.

EDIT2: Nvm, I am stupid. This always trips me up

<del>
Also found an error by upgrading this.

https://github.com/davidcole1340/ext-php-rs/blob/5a02863c3b672e44353ba9aca6f22776a9386b93/src/zend/globals.rs#L205-L211
and
https://github.com/davidcole1340/ext-php-rs/blob/5a02863c3b672e44353ba9aca6f22776a9386b93/src/zend/globals.rs#L218-L224

conditions are the wrong way around if I see it correctly. Verifying right now. 

EDIT: Not the wrong way around, just not the right condition

Issue #481, opened 2025-07-03
Consider escalating priviledge to `sudo` automatically when copying `.so` file or editing ini file


dilawar:
https://github.com/davidcole1340/ext-php-rs/blob/060c8bfb2dacffe35cffb558d4f3738d86a5d34c/crates/cli/src/lib.rs#L224 

The problem 

If you are outside docker and/or don't have permissions to write into php's config directory.

- `cargo php install` will fails if normal user don't have permissions to copy file or write to `.ini` file.
- `sudo cargo php install` is likely to fail since `cargo` is installed in user's `~/.cargo/bin` directory.
- `sudo ~/.cargo/bin/cargo php install` is likely to rebuild and change the owner to `root` of many intermediate files make normal `cargo b` to fail.

Solution: before copying file, check it it requires admin privilge and use `sudo2` or `elevate` crate to escalate priviledge.

dilawar:
Would you accept a MR if this make sense?

Xenira:
I really like that idea. I would also like to add a root check and just fail unless something like `--bypass-root-check` is passed.

dilawar:
Getting root check on Windows may be tricky. I am not even sure how PHP installation works on Windows.

PR #482, opened 2025-07-03
feat(cargo-php)!: escalate privilege and to copy extension and edit ini file

dilawar:
Fixes: #481
BREAKING CHANGE: installing extensions as root user will now fail unless `--bypass-root-check` is provided

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -19,7 +19,9 @@ dialoguer = "0.11"
 libloading = "0.8"
 cargo_metadata = "0.20"
 semver = "1.0"
-elevate = "0.6.1"
+
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
 
 [lints.rust]
 missing_docs = "warn"
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -142,6 +142,7 @@ struct Remove {
     /// Whether to bypass the remove prompt.
     #[clap(long)]
     yes: bool,
+    #[cfg(unix)]
     /// Whether to bypass the root check
     #[clap(long)]
     bypass_root_check: bool,
@@ -198,12 +199,11 @@ impl Install {
             self.no_default_features,
         )?;
 
-        if !self.bypass_root_check {
-            anyhow::ensure!(
-                elevate::check() == elevate::RunningAs::User,
-                "Running as root is not recommended. Use --bypass-root-check to override."
-            );
-        }
+        #[cfg(unix)]
+        anyhow::ensure!(
+            self.bypass_root_check || !is_root(),
+            "Running as root is not recommended. Use --bypass-root-check to override."
+        );
 
         let (mut ext_dir, mut php_ini) = if let Some(install_dir) = self.install_dir {
             (install_dir, None)
@@ -246,10 +246,10 @@ impl Install {
     }
 }
 
-// Update extension information in the ini file, if fails, try with sudo again.
-//
-// Write to a temp file then move it to given path. Use `sudo` on unix to move
-// file if needed.
+/// Update extension line in the ini file.
+///
+/// Write to a temp file then copy it to a given path. If this fails, then try
+/// `sudo mv` on unix.
 fn update_ini_file(php_ini: &PathBuf, ext_name: &str, disable: bool) -> anyhow::Result<()> {
     let current_ini_content = std::fs::read_to_string(php_ini)?;
     let mut ext_line = format!("extension={ext_name}");
@@ -267,25 +267,29 @@ fn update_ini_file(php_ini: &PathBuf, ext_name: &str, disable: bool) -> anyhow::
     }
 
     new_lines.push(&ext_line);
+
     write_to_file(new_lines.join("\n"), php_ini)?;
     Ok(())
 }
 
-// Copy extension, if fails, try with sudo cp.
-//
-// Checking if we have write permission for ext_dir may fail due to ACL, group
-// list and and other nuances. See
-// https://doc.rust-lang.org/std/fs/struct.Permissions.html#method.readonly
+/// Copy extension, if fails, try with sudo cp.
+///
+/// Checking if we have write permission for ext_dir may fail due to ACL, group
+/// list and and other nuances. See
+/// https://doc.rust-lang.org/std/fs/struct.Permissions.html#method.readonly.
 fn copy_extension(ext_path: &Utf8PathBuf, ext_dir: &PathBuf) -> anyhow::Result<()> {
     if let Err(_e) = std::fs::copy(ext_path, ext_dir) {
         #[cfg(unix)]
         {
-            let _ = std::process::Command::new("sudo")
+            let s = std::process::Command::new("sudo")
                 .arg("cp")
                 .arg(ext_path)
                 .arg(ext_dir)
                 .status()?;
+            anyhow::ensure!(s.success(), "Failed to copy extension");
         }
+        #[cfg(not(unix))]
+        anyhow::bail!("Failed to copy extension: {_e}");
     }
     Ok(())
 }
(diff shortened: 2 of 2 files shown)


crates/cli/Cargo.toml
```diff
@@ -19,6 +19,7 @@ dialoguer = "0.11"
 libloading = "0.8"
 cargo_metadata = "0.20"
 semver = "1.0"
+sudo = "0.6.0"
```
Xenira: This crate is no longer maintained afaik.

crates/cli/src/lib.rs
```diff
@@ -193,6 +199,10 @@ impl Install {
             self.no_default_features,
         )?;
 
+        if !self.bypass_root_check {
+            escalate_root();
```
Xenira: Not what I had in mind.

Was looking for [privilege seperation](https://en.m.wikipedia.org/wiki/Privilege_separation).

When doing `sudo cargo php install` we should fail instantly with a message like 'this should not be executed as root. If you know what you are doing execute it again with the --bypass-root-check flag'.

Only the copy operation should be executed with root privileges.
As this is a simple `cp` operation I don't think we need to overcomplicate this and can just use `Command` to do this and guard that for *nix systems.

Xenira (changes requested):

crates/cli/Cargo.toml
dilawar: There are two forks that I know of, but they both pull extra deps 'tracing'. Since this project doesn't use tracing, I preferred to use this one.

crates/cli/src/lib.rs
dilawar: I totally misunderstood you. This is better and doesn't even need `sudo` crate.

crates/cli/src/lib.rs
```diff
@@ -221,45 +233,63 @@ impl Install {
...
+fn copy_extension(ext_path: &Utf8PathBuf, ext_dir: &PathBuf) -> anyhow::Result<()> {
+    if let Err(_e) = std::fs::copy(ext_path, ext_dir) {
+        #[cfg(unix)]
+        {
+            let _ = std::process::Command::new("sudo")
+                .arg("cp")
+                .arg(ext_path)
+                .arg(ext_dir)
+                .status()?;
+        }
```
Xenira: This silently drops the error on non unix systems

crates/cli/src/lib.rs
```diff
@@ -568,3 +598,34 @@ fn build_ext(
...
+            let _ = std::process::Command::new("sudo")
+                .arg("mv")
+                .arg(&tempf)
+                .arg(filepath)
+                .status()?;
+        }
+
+        #[cfg(not(unix))]
+        anyhow::bail!("failed to write to {filepath:?}");
+    }
```
Xenira: Technically this could override changes made between the time the tmp file is created and its copied.

But I think that is fine for now.

Xenira (changes requested):
That looks a lot like what I was thinking about. Thanks!

Maybe we can check if we are running with root without needing the dependency, but haven't looked into that.

crates/cli/src/lib.rs
dilawar: Missed that! Will make changes.

dilawar:
> Maybe we can check if we are running with root without needing the dependency, but haven't looked into that.

I'll borrow the function from `elevate/sudo` crate that check if we are running as root. Knowing the `uid`/`gui` should be sufficient here.

crates/cli/src/lib.rs
dilawar: I tried adding a write lock, but then I didn't know how to do it properly on Windows without pulling-in `windows-rs` crate.

There is https://crates.io/crates/tempfile that may be a cross-platform solution.

crates/cli/Cargo.toml
```diff
@@ -20,5 +20,8 @@ libloading = "0.8"
 cargo_metadata = "0.20"
 semver = "1.0"
 
+[target.'cfg(unix)'.dependencies]
```
dilawar, line 23: Using it to determine if current user is `root`/`sudo` or not! This crate is already in `Cargo.lock` and at least used by two other deps.

crates/cli/src/lib.rs
```diff
@@ -193,6 +198,14 @@ impl Install {
             self.no_default_features,
         )?;
 
+        #[cfg(unix)]
+        if !self.bypass_root_check {
+            anyhow::ensure!(
+                !is_root(),
+                "Running as root is not recommended. Use --bypass-root-check to override."
+            );
+        }
```
Xenira:
```suggestion
        anyhow::ensure!(
            self.bypass_root_check || !is_root(),
            "Running as root is not recommended. Use --bypass-root-check to override."
        );
```

crates/cli/src/lib.rs
```diff
@@ -221,45 +234,67 @@ impl Install {
...
+
+    write_to_file(new_lines.join("\n"), php_ini)?;
+    Ok(())
+}
+
+// Copy extension, if fails, try with sudo cp.
+//
+// Checking if we have write permission for ext_dir may fail due to ACL, group
+// list and and other nuances. See
+// https://doc.rust-lang.org/std/fs/struct.Permissions.html#method.readonly.
```
Xenira:
```suggestion
/// Copy extension, if fails, try with sudo cp.
///
/// Checking if we have write permission for ext_dir may fail due to ACL, group
/// list and and other nuances. See
/// https://doc.rust-lang.org/std/fs/struct.Permissions.html#method.readonly.
```

crates/cli/src/lib.rs
```diff
@@ -568,3 +603,46 @@ fn build_ext(
 
     bail!("Failed to retrieve extension path from artifact")
 }
+
+// Write content to a given filepath.
+//
+// We may not have write permission but we may have sudo privilege on unix. So
+// we write to a temp file and then try moving it to given filepath, and retry
+// with sudo on unix.
```
Xenira:
```suggestion
/// Write content to a given filepath.
///
/// We may not have write permission but we may have sudo privilege on unix. So
/// we write to a temp file and then try moving it to given filepath, and retry
/// with sudo on unix.
```

crates/cli/src/lib.rs
```diff
@@ -568,3 +603,46 @@ fn build_ext(
...
+        #[cfg(not(unix))]
+        anyhow::bail!("failed to write to {filepath:?}");
+    }
+
+    Ok(())
+}
+
+#[cfg(unix)]
+fn is_root() -> bool {
+    // nix::unistd::Uid::effective().is_root()
```
Xenira:
```suggestion
```

crates/cli/src/lib.rs
```diff
@@ -568,3 +603,46 @@ fn build_ext(
...
+    let uid = unsafe { libc::getuid() };
+    let euid = unsafe { libc::geteuid() };
+
+    match (uid, euid) {
+        (0, 0) => true,
+        (_, 0) => true, // suid set
```
Xenira: The second case always matches if the first one does.

This would be equivalent:

```suggestion
        (_, 0) => true, // suid set
```

crates/cli/src/lib.rs
```diff
@@ -221,45 +234,67 @@ impl Install {
...
         }
 
         Ok(())
     }
 }
 
+// Update extension line in the ini file.
+//
+// Write to a temp file then copy it to a given path. If this fails, then try
+// `sudo mv` on unix.
```
Xenira:
```suggestion
/// Update extension line in the ini file.
///
/// Write to a temp file then copy it to a given path. If this fails, then try
/// `sudo mv` on unix.
```

crates/cli/src/lib.rs
```diff
@@ -140,6 +142,9 @@ struct Remove {
     /// Whether to bypass the remove prompt.
     #[clap(long)]
     yes: bool,
+    /// Whether to bypass the root check
+    #[clap(long)]
+    bypass_root_check: bool,
```
Xenira, line 148: Only need that option on unix systems

```suggestion
    #[cfg(unix)]
    /// Whether to bypass the root check
    #[clap(long)]
    bypass_root_check: bool,
```

Xenira (changes requested):

Xenira:
@dilawar Thank you! Could you update the guide documentation to include the new behaviour?

The git commit history could benefit from a cleanup as well. See the [CONTRIBUTING.md](https://github.com/davidcole1340/ext-php-rs/blob/master/CONTRIBUTING.md#commit-messages).

dilawar:
I'll definitely give it a shot, but I am terrible at writing!  I'll rebase/cleanup git history.

Xenira:
> I'll definitely give it a shot, but I am terrible at writing! I'll rebase/cleanup git history.

I can do the documentation :)

kakserpom:
[nix::unistd](https://docs.rs/nix/latest/nix/unistd/index.html) looks like a better fit.

Issue #455, opened 2025-06-17
Add issue templates to improve issue quality


Xenira:
<!-- see this issue why this is needed

PR #483, opened 2025-07-03
chore(github): add issue and pr templates

Xenira:
Refs: #455


Xenira:
@ptondereau not sure if they are to involved. Want to have some structure and collect the required information while not overwhelming ppl.

That is one reason why i'd still allow plain issue creation.

What do you think?

ptondereau (approved):

ptondereau:
> @ptondereau not sure if they are to involved. Want to have some structure and collect the required information while not overwhelming ppl.
> 
> That is one reason why i'd still allow plain issue creation.
> 
> What do you think?

It's ok to me. 
For bug report the main concern I faced is always asking the PHP version and the lib version 😄

PR #488, opened 2025-07-06
feat: make Sapi work with ZTS builds

Qard:
## Description

This enables a `SapiModule` to work in ZTS mode. I'm using `SapiModule` directly in my case rather than going through the `Embed` type. I missed this when submitting my other PR to expand `SapiBuilder`.

Not sure what making `Embed` ZTS-capable would look like as it would require constructing a single `SapiModule` and reusing it so the API might need to be different or it might need to do some internal `OnceCell` magic.

Not entirely sure how to test this in a way that's not rather complicated. The only observable difference I'm aware of is that without the change it _might_ randomly segfault eventually with enough parallel requests. Any ideas? 😅 

I'm already using this code in my fork, so I know it _works_, just writing a test that's not _gigantic_ would be a challenge. 🤔 

## Checklist

### DIFF
--- a/src/embed/embed.c
+++ b/src/embed/embed.c
@@ -29,6 +29,21 @@ void ext_php_rs_sapi_startup() {
   zend_signal_startup();
 }
 
+SAPI_API void ext_php_rs_sapi_shutdown() {
+  #ifdef ZTS
+    tsrm_shutdown();
+  #endif
+}
+
+SAPI_API void ext_php_rs_sapi_per_thread_init() {
+  #ifdef ZTS
+    (void)ts_resource(0);
+    #ifdef PHP_WIN32
+      ZEND_TSRMLS_CACHE_UPDATE();
+    #endif
+  #endif
+}
+
 void ext_php_rs_php_error(int type, const char *format, ...) {
   va_list args;
   va_start(args, format);
--- a/src/embed/embed.h
+++ b/src/embed/embed.h
@@ -7,5 +7,7 @@
 void* ext_php_rs_embed_callback(int argc, char** argv, void* (*callback)(void *), void *ctx);
 
 void ext_php_rs_sapi_startup();
+void ext_php_rs_sapi_shutdown();
+void ext_php_rs_sapi_per_thread_init();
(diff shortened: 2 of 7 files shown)


Xenira:
> Not entirely sure how to test this in a way that's not rather complicated. The only observable difference I'm aware of is that without the change it _might_ randomly segfault eventually with enough parallel requests. Any ideas? 😅
> 
> I'm already using this code in my fork, so I know it _works_, just writing a test that's not _gigantic_ would be a challenge. 🤔

Not having tests here would be fine here. It is not a core functionality and the effort is not justified. Should this break at any time we might want to consider doing it anyways, but for now I am ok with not testing.

Qard:
I actually was able to put together a very _simple_ test for multi-threaded SAPI. I misremembered a bit about what was needed for `Embed`, so you can actually use _that_ in a multi-threaded environment too. 🙂

Qard:
Ah, the test seems to be a bit problematic. It runs fine with `cargo test --test sapi_tests --features embed`, but if you turn on the `closure` feature (which CI is doing) it explodes because apparently you can only call `get_module` _once_ as it appears the `Closure::build()` initialization  doesn't clean up after itself when the module is torn down.

Possibly best to just delete that test for now, unless you know a good way to get that to clean up properly? 🤔

Xenira:
> Ah, the test seems to be a bit problematic. It runs fine with `cargo test --test sapi_tests --features embed`, but if you turn on the `closure` feature (which CI is doing) it explodes because apparently you can only call `get_module` _once_ as it appears the `Closure::build()` initialization doesn't clean up after itself when the module is torn down.
> 
> Possibly best to just delete that test for now, unless you know a good way to get that to clean up properly? 🤔

Could you extract the test case into a separate MR? That way we don't loose track of it and can check if there is a good way of cleaning up. Would be ok with merging this without the testcase being included.

Qard:
Sure, I'll do that today.

Qard:
The test has been moved to #496. I'll rebase that and continue after this lands.

Issue #178, opened 2022-11-10
Php enums


estk:
Hi @davidcole1340,
I'm wondering why php 8 enums were not implemented, is there some issue with defining an enum in an extension?
Asking because I am currently working on it and would be interested in upstreaming the change.

davidcole1340:
No reason -- I just haven't had time to do it. PRs are greatly appreciated if you are working on it!

Some references if you would like to have a look:
- https://github.com/php/php-src/blob/f6c9a90e19d648355c39da84af9bc59b2ff60c15/Zend/zend_enum.c#L480-L509
- https://github.com/php/php-src/blob/f6c9a90e19d648355c39da84af9bc59b2ff60c15/ext/zend_test/test_arginfo.h#L575-L636

estk:
Excellent, thanks @davidcole1340. I should have a PR for you shortly. Also, thanks for the refs, especially the test ref, I didn't know that existed.

joehoyle:
Given enums were added in PHP 8.1, I'm not sure the best way to handle the optionality of them. Should this maybe be a cargo feature?

danog:
I'd rather just bump the minimum version, since PHP 8.0 has already stopped receiving security updates.

Xenira:
@joehoyle I know that we have spoken about this, but was not aware of this at that time.

Not sure how hard this would break 8.0, but I'd rather remove support entirely then have it maybe work.
If someone really needs to compile against an old php version they can still build from src and tinker to get it working.

As long as we are upfront on how long a given version is supported we should be fine removing old ones.
This not only allows us to add newer features more easily, but reduces maintainance.

PR #489, opened 2025-07-06
feat(enum): add basic enum support

Xenira:
Fixes: #178
Refs: #302

## Description

PHP 8.1 introduced enums. This is a migration of #302 to the new macro system.

MR is currently incomplete, but the basic case already works.

I feature guarded this behind the `enum` feature. As I added this to the default features this is a soft bump of the minimum supported PHP version. Setting `default-features` to false will still allow 8.0 though.

### Progress

- [x] Register unit enums
- [x] Register backed enums
- [x] Enums as return types (`IntoZval`)
- [x] Enums as parameters (`FromZval`)
- [x] `TryFrom` for backed enums
- [x] Stubs
- [x] Tests
- [x] Documentation

## Checklist

- [x] I have added a [migration guide](../CONTRIBUTING.md#breaking-changes) if applicable.


Xenira:
@ptondereau thoughts?

Want to finalise the api before tackling the test.

src/builders/enum_builder.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        for case in self.cases {
+            let name = ZendStr::new_interned(case.name, true);
+            let value = match &case.discriminant {
+                Some(value) => {
+                    let value: Zval = match value {
+                        crate::enum_::Discriminant::Int(i) => i.into_zval(false)?,
+                        crate::enum_::Discriminant::String(s) => s.into_zval(true)?,
+                    };
+                    let mut zv = core::mem::ManuallyDrop::new(value);
```
ptondereau: We should implement the RAII box pattern then like:

```rust
fn create_enum_value(discriminant: &Discriminant) -> Result<*mut zval> {
    let value: Zval = match discriminant {
        Discriminant::Int(i) => i.into_zval(false)?,
        Discriminant::String(s) => s.into_zval(true)?,
    };

    let boxed_value = Box::new(value);
    Ok(Box::into_raw(boxed_value).cast())
}

// ...
let value_ptr = create_enum_value(value)?;
unsafe {
    zend_enum_add_case(class, name.into_raw(), value_ptr);
}
 
```

crates/macros/src/enum_.rs
```diff
@@ -0,0 +1,411 @@
...
+        quote! {
+            impl ::ext_php_rs::enum_::RegisteredEnum for #ident {
+                const CASES: &'static [::ext_php_rs::enum_::EnumCase] = &[
+                    #(#cases,)*
+                ];
+
+                fn from_name(name: &str) -> ::ext_php_rs::error::Result<Self> {
+                    match name {
+                        #(#case_from_names,)*
+                        _ => Err(::ext_php_rs::error::Error::InvalidProperty),
```
ptondereau, line 240: this is sad that InvalidProperty doesn't take some metadata, because at this point we cannot know from macro which case fails. But can do it later

src/builders/enum_builder.rs
```diff
@@ -0,0 +1,210 @@
...
+        let class = unsafe {
+            zend_register_internal_enum(
+                CString::new(self.name)?.as_ptr(),
+                self.datatype.as_u32().try_into()?,
+                methods.into_boxed_slice().as_ptr(),
+            )
+        };
+
+        for case in self.cases {
+            let name = ZendStr::new_interned(case.name, true);
```
ptondereau, line 115: this is a good first approach, but I think with a large backed enum, it will load a lot of allocation.
Maybe consider later a string pool generated by a macro or string arena?

ptondereau (approved):
This is a good feature! Thanks for the implementation. 
Leaved some nitpicks.

src/builders/enum_builder.rs
Xenira: From what I can see it is converted into an interned string if it isn't already on php side. Probably only worth investigating if we notice real world impact.

https://github.com/php/php-src/blob/4647dfc9887c5d506facd4d559500193484efd1c/Zend/zend_enum.c#L581-L583

Xenira: Nvm, thats the name...

ptondereau: ok I guess it is a PHP problem then 😅

crates/macros/src/enum_.rs
Xenira: Yeah, error types are not optimal right now.

Think a bigger overhaul is in order here. Using `thiserror` with more specialised errors could be helpful.

Kinda tempted to leave it like this and do a bigger error handling MR.

ptondereau (approved):

src/builders/enum_builder.rs
Xenira: Ok, name can be not interned. Will change that. Only value seems to be interned if its not.

Xenira: @ptondereau any idea why non interned string fails on macos nts?

Xenira: Kinda remember that I had that problem in a previous iteration also on mac.

ptondereau: I don't have a clew 😢

Xenira: Well, then Ill merge it with interned and have a look at it if we notice problems using interned strings.

joehoyle:
Great stuff thanks @Xenira

PR #496, opened 2025-07-10
Add sapi zts test

Qard:
## Description

Depends on #488.

I've pulled the test for that out as it had a problematic interaction with the `closure` feature. I'll have to figure out how to make that work properly. 🤔 

## Checklist


ptondereau:
Hello @Qard 
We think your PR is a good addition.
Could it be possible to rebase here?

Qard:
Yep, I can try a rebase. Though I never did get around to looking into the closure interaction. It may still be broken?

Qard:
I've pushed a possible fix to the closure conflict, though I'm uncertain if that is actually correct. Anyone else know the closure subsystem better that can review that?

ptondereau:
> I've pushed a possible fix to the closure conflict, though I'm uncertain if that is actually correct. Anyone else know the closure subsystem better that can review that?

I'll try to have a look next week. Thanks for the work

Qard:
> > I've pushed a possible fix to the closure conflict, though I'm uncertain if that is actually correct. Anyone else know the closure subsystem better that can review that?
> 
> I'll try to have a look next week. Thanks for the work

It seemed to get past the original failure, but now fails in a different. Unfortunately though it seems to only fail in CI for me and not locally, so I'm not too sure how to debug. If I get time at some point I might take another look, but I'm not too sure how much sense I'd be able to make of it. 🤔

ptondereau:
> > > I've pushed a possible fix to the closure conflict, though I'm uncertain if that is actually correct. Anyone else know the closure subsystem better that can review that?
> > 
> > I'll try to have a look next week. Thanks for the work
> 
> It seemed to get past the original failure, but now fails in a different. Unfortunately though it seems to only fail in CI for me and not locally, so I'm not too sure how to debug. If I get time at some point I might take another look, but I'm not too sure how much sense I'd be able to make of it. 🤔 

Embed test runs only on 8.4 (embed sapi) debug nts with latest rust and llvm 17.

ptondereau:
Ok, this is because I guess you have a ZTS build of PHP locally, and the Test embed runs only on a non-TS version.
This means that you probably need to add a test guard for ZTS build (we already have a `#[cfg(php_zts)]`) and add a matrix entry for test embed with PHP ZTS/NTS.

Qard:
Where exactly? I'm not clear exactly what you're suggesting I change. 🤔

tests/sapi.rs
```diff
@@ -102,3 +117,91 @@ pub fn hello_world(name: String) -> String {
 pub fn module(module: ModuleBuilder) -> ModuleBuilder {
     module.function(wrap_function!(hello_world))
 }
+
+#[test]
```
ptondereau, line 125:
```suggestion
#[test]
#[cfg(php_zts)]
```

Qard: Ah, you meant skipping it altogether. Makes sense. Thanks for the guidance!

ptondereau: Yes and you can add to test-embed ci job, a ts entry to run this job against PHP with zts

ptondereau:
Now that the embed tests pass, we can introduce ZTS with a matrix test for `test-embed` job: https://github.com/extphprs/ext-php-rs/blob/master/.github/workflows/build.yml#L172 
I can add a commit in your PR if you'd like.

Qard:
Sure. I probably won't have the time to look at it this week. 😅

ptondereau:
I struggle to push on your branch but you can see the diff here: https://github.com/extphprs/ext-php-rs/pull/589/files

.github/workflows/build.yml
ptondereau: could you try removing this step?

Qard: Tried that. Seems to build now, but fails with a SIGSEGV. Unfortunately I don't have the time to debug that at the moment so I'll have to pause this again for now. 🤔

Qard:
Figured out the failure. It was missing a thread cleanup which was crashing the mongodb extension on shutdown.

ptondereau:
> Figured out the failure. It was missing a thread cleanup which was crashing the mongodb extension on shutdown.

Awesome thank you very much

Qard:
@ptondereau Anything else you need for this, or can this be landed?

ptondereau:
> @ptondereau Anything else you need for this, or can this be landed?

Can you make a rebase just make sure. Thanks

Qard:
Done. 🙂

Issue #477, opened 2025-07-02
Zval::null() to get a null in one call


kakserpom:
I wrote this trait some time ago:

```rust
pub trait ZvalNull {
    /// Returns a new `Zval` explicitly set to `null`.
    fn null() -> Zval;
}

impl ZvalNull for Zval {
    fn null() -> Zval {
        let mut zval = Zval::new();
        zval.set_null();
        zval
    }
}
```

What do you guys think about adding such method to `impl Zval`?

Xenira:
Could be helpful. Might want to check if there is already a const in zend that can be used. Like #355

kakserpom:
Done in https://github.com/davidcole1340/ext-php-rs/pull/521

PR #520, opened 2025-07-16
Array btreemap conversion

kakserpom:
https://github.com/davidcole1340/ext-php-rs/issues/477

### DIFF
--- a/src/types/array.rs
+++ b/src/types/array.rs
@@ -1,16 +1,6 @@
 //! Represents an array in PHP. As all arrays in PHP are associative arrays,
 //! they are represented by hash tables.
 
-use std::{
-    collections::HashMap,
-    convert::{TryFrom, TryInto},
-    ffi::CString,
-    fmt::{Debug, Display},
-    iter::FromIterator,
-    ptr,
-    str::FromStr,
-};
-
 use crate::{
     boxed::{ZBox, ZBoxable},
     convert::{FromZval, IntoZval},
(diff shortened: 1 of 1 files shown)


Xenira:
@kakserpom thank you <3

Just some things:

1. Could you please use the MR template. It is there for a reason. No need to add a description when referencing an issue, but it serves as a reminder for both the author and the reviewer on what to check.
2. Could you split this in 2 MRs, as these are 2 unrelated features
3. Unit and integration tests are missing.
4. At least the `BTreeMap` would benefit from being mentioned in the guides
5. `cargo fmt` fails. We have git hooks that could catch that if you want to use them :)

kakserpom:
@Xenira Sure, now it's split into two MRs. I don't believe there's anything to test.

src/types/array.rs
```diff
@@ -1288,6 +1287,76 @@ where
...
 }
 
+///////////////////////////////////////////
+// BTreeMap
+///////////////////////////////////////////
+
+impl<'a, V> TryFrom<&'a ZendHashTable> for BTreeMap<String, V>
+where
+    V: FromZval<'a>,
+{
```
Xenira, line 1297: Can we do this here? Same for the others.

```suggestion
impl<'a, K, V> TryFrom<&'a ZendHashTable> for BTreeMap<K, V>
where
    K: TryFrom<ArrayKey<'a>, Error = Error>,
    V: FromZval<'a>,
{
```

ptondereau:
hey! we split array.rs into smaller files. You can rebase and create a dedicated `src/types/array/conversions/btreemap.rs`

src/types/array.rs
kakserpom: @Xenira Not sure this is a good idea.

kakserpom:
https://github.com/davidcole1340/ext-php-rs/pull/535

PR #529, opened 2025-07-19
refactor(array): split array.rs types into smaller files

ptondereau:
## Description

A dummy split only here, but not sure about naming.


src/util/mod.rs
```diff
@@ -1,3 +1 @@
-mod cstring_scope;
-
-pub use cstring_scope::CStringScope;
```
ptondereau, line 3: It creates a warning that it's unused. I don't know if it's the best way to handle this.

Xenira:
Thanks. This is pretty much what I had in mind. Would further split the conversions though, as I see them growing in the future.

Something like
```
hash_map.rs
vec.rs
btree_map.rs
hash_set.rs
indexmap.rs
[...]
```

src/util/mod.rs
Xenira: Ye, not sure either. Prob best to just add an `#[allow]` for now.

ptondereau:
> Thanks. This is pretty much what I had in mind. Would further split the conversions though, as I see them growing in the future.
> 
> Something like
> 
> ```
> hash_map.rs
> vec.rs
> btree_map.rs
> hash_set.rs
> indexmap.rs
> [...]
> ```

I've created a `conversions` sub module and I don't know if `from_iterator.rs` should be part of `iterator.rs` or should be living alone.

Xenira:
I would keep it in the `iterator.rs` for now. Don't see it growing that much rn and it fits better in there.

Xenira (approved):

Xenira:
Thanks, looks good!

Xenira (changes requested):
Ohh, just noticed. Mind moving the tests into the vec conversion module?

ptondereau:
> Ohh, just noticed. Mind moving the tests into the vec conversion module?

I split the tests, sorry.

src/types/array/array_key.rs
```diff
@@ -0,0 +1,155 @@
...
+        if let Some(key) = zval.string() {
+            return Some(ArrayKey::String(key));
+        }
+        None
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+mod tests {
```
Xenira: These need embed as well, not entirely sure why but they somehow depend on php types...

PR #535, opened 2025-07-20
feat(array): BTreeMap conversion

kakserpom:
## Description

I've added the same `FromZval`/`IntoZval` conversions for `BTreeMap` that exist for `HashMap`.

I know a doc update wouldn't hurt, but I am too preoccupied with development atm.

### DIFF
--- a/src/types/array/conversions/mod.rs
+++ b/src/types/array/conversions/mod.rs
@@ -9,5 +9,6 @@
 //! - `HashMap<K, V>` ↔ `ZendHashTable` (via `hash_map` module)
 //! - `Vec<T>` and `Vec<(K, V)>` ↔ `ZendHashTable` (via `vec` module)
 
+mod btree_map;
 mod hash_map;
 mod vec;
--- a/src/types/array/conversions/btree_map.rs
+++ b/src/types/array/conversions/btree_map.rs
@@ -1,0 +1,77 @@
+use std::{collections::BTreeMap, convert::TryFrom};
+
+use crate::{
+    boxed::ZBox,
+    convert::{FromZval, IntoZval},
+    error::{Error, Result},
+    flags::DataType,
+    types::Zval,
+};
+
+use super::super::ZendHashTable;
+
+impl<'a, V> TryFrom<&'a ZendHashTable> for BTreeMap<String, V>
+where
+    V: FromZval<'a>,
+{
+    type Error = Error;
+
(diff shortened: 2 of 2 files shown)


src/types/array/conversions/mod.rs
```diff
@@ -9,5 +9,6 @@
 //! - `HashMap<K, V>` ↔ `ZendHashTable` (via `hash_map` module)
 //! - `Vec<T>` and `Vec<(K, V)>` ↔ `ZendHashTable` (via `vec` module)
```
ptondereau, line 11: Could you please add some docs here?

Xenira:
Also could you elaborate on this? https://github.com/davidcole1340/ext-php-rs/pull/520#discussion_r2217875652

kakserpom:
@Xenira 

> Also could you elaborate on this? [#520 (comment)](https://github.com/davidcole1340/ext-php-rs/pull/520#discussion_r2217875652)

I've implemented your suggestion. I've also fixed the cause of `clippy::implicit_hasher` warning in `hash_map` by generalizing the hasher.

src/types/array/conversions/mod.rs
kakserpom: Done.

kakserpom:
> LGTM but would it be possible to have some tests?

What's there to test? Rustc does the job pretty well :)

ptondereau:
> > LGTM but would it be possible to have some tests?
> 
> What's there to test? Rustc does the job pretty well :)

Just ensuring the conversion if it makes sense

kakserpom:
I've changed the behavior of `impl TryFrom<ArrayKey<'_>> for String`. Now it converts numeric keys into strings, instead of giving an error.

kakserpom:
And also I've moved the code that treats numeric string keys as` long`, now it happens in ArrayKey conversion impls. So no more boilerplate in `impl ZendHashTable`.

Xenira:
> I've changed the behavior of `impl TryFrom<ArrayKey<'_>> for String`. Now it converts numeric keys into strings, instead of giving an error.

You are probably right. This would indeed be the desired behaviour for `TryFrom`.

> And also I've moved the code that treats numeric string keys as` long`, now it happens in ArrayKey conversion impls. So no more boilerplate in `impl ZendHashTable`.

This has some implications on types and I don't like how much we implicitly convert string and int values. But that is an issue for later.

Issue #541, opened 2025-08-04
Add visability to constructor


Norbytus:
## Description
For now all constructor generate as public, and not possibility to declare as private or protected

## Use Case
Some time need declare class __construct as private or protected, and instead construct use
```php
public static function new()
```
or to escape from error ```You cannot instantiate this class from PHP.```

## Example
```rust
#[php_class]
struct TestClassMethodVisability;

#[php_impl]
impl TestClassMethodVisability {
    #[php(vis = "private")]
    fn __construct() -> Self {
        Self
    }

    #[php(vis = "private")]
    fn private() -> u32 {
        3
    }

    #[php(vis = "protected")]
    fn protected() -> u32 {
        3
    }
}
#[php_class]
struct TestClassProtectedConstruct;

#[php_impl]
impl TestClassProtectedConstruct {
    #[php(vis = "protected")]
    fn __construct() -> Self {
        Self
    }
}
```

```php
#[php_class]
struct TestClassMethodVisability;

#[php_impl]
impl TestClassMethodVisability {
    #[php(vis = "private")]
    fn __construct() -> Self {
        Self
    }

    #[php(vis = "private")]
    fn private() -> u32 {
        3
    }

    #[php(vis = "protected")]
    fn protected() -> u32 {
        3
    }
}
#[php_class]
struct TestClassProtectedConstruct;

#[php_impl]
impl TestClassProtectedConstruct {
    #[php(vis = "protected")]
    fn __construct() -> Self {
        Self
    }
}
```

### Additional Context

- [x] I can try implementing this feature myself (please assign me to this issue)

PR #542, opened 2025-08-04
feat: Add constructor visability

Norbytus:
Add visability in constructor meta, and feature mark constructor as private and protected

## Description

- Add MethodFlags to ConstructorMeta 
- Relevant or closing issues[ (https://github.com/davidcole1340/ext-php-rs/issues/541)
- I made fast solution, and it could be not neat

## Checklist

- [x] I have added a [migration guide](../CONTRIBUTING.md#breaking-changes) if applicable.

### DIFF
--- a/crates/macros/src/lib.rs
+++ b/crates/macros/src/lib.rs
@@ -712,8 +712,8 @@ fn php_module_internal(args: TokenStream2, input: TokenStream2) -> TokenStream2 
 /// - `#[php(optional = i)]` - Sets the first optional parameter. Note that this
 ///   also sets the remaining parameters as optional, so all optional parameters
 ///   must be a variant of `Option<T>`.
-/// - `#[php(public)]`, `#[php(protected)]` and `#[php(private)]` - Sets the
-///   visibility of the method.
+/// - `#[php(vis = "public")]`, `#[php(vis = "protected")]` and `#[php(vis =
+///   "private")]` - Sets the visibility of the method.
 /// - `#[php(name = "method_name")]` - Renames the PHP method to a different
 ///   identifier, without renaming the Rust method name.
 ///
--- a/src/builders/class.rs
+++ b/src/builders/class.rs
@@ -238,16 +238,21 @@ impl ClassBuilder {
             "Class name in builder does not match class name in `impl RegisteredClass`."
         );
         self.object_override = Some(create_object::<T>);
-        self.method(
-            {
-                let mut func = FunctionBuilder::new("__construct", constructor::<T>);
-                if let Some(ConstructorMeta { build_fn, .. }) = T::constructor() {
-                    func = build_fn(func);
-                }
-                func
-            },
-            MethodFlags::Public,
-        )
+
+        let (func, visibility) = if let Some(ConstructorMeta {
+            build_fn, flags, ..
+        }) = T::constructor()
+        {
+            let func = FunctionBuilder::new("__construct", constructor::<T>);
+            (build_fn(func), flags.unwrap_or(MethodFlags::Public))
+        } else {
+            (
+                FunctionBuilder::new("__construct", constructor::<T>),
+                MethodFlags::Public,
+            )
+        };
+
+        self.method(func, visibility)
     }
 
     /// Function to register the class with PHP. This function is called after
--- a/src/class.rs
+++ b/src/class.rs
@@ -82,6 +82,8 @@ pub struct ConstructorMeta<T> {
     /// Function called to build the constructor function. Usually adds
     /// arguments.
     pub build_fn: fn(FunctionBuilder) -> FunctionBuilder,
+    /// Add constructor modification
+    pub flags: Option<MethodFlags>,
 }
 
 /// Result returned from a constructor of a class.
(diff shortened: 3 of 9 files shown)


Norbytus:
@Xenira  how i can update guide there wrong doc about private,protected, public
Now 
```
#[php(public)]`, `#[php(protected)]` and `#[php(private)]` - Sets the
```

Should be
```
#[php(vis = "public")]`, `#[php(vis = "protected")]` and `#[php(vis = "private")]` - Sets the
```

Xenira:
@Norbytus see here: https://github.com/davidcole1340/ext-php-rs/blob/34caede39f58204f11afec767f3b6ddd6648c937/guide/src/macros/impl.md#L45

Norbytus:
> @Norbytus see here:
> 
> https://github.com/davidcole1340/ext-php-rs/blob/34caede39f58204f11afec767f3b6ddd6648c937/guide/src/macros/impl.md#L45

I see that variant, but it's not working)
```
error: Unknown field: `private`
   --> tests/src/integration/class/mod.rs:152:11
    |
152 |     #[php(private)]
    |           ^^^^^^^
```
 
So i think, maybe we update dock.
But anyway, main thing for this PR it's declare __construct as public, private, protected

Xenira:
Ye, docs are probably wrong there.

crates/macros/src/lib.rs
```diff
@@ -712,7 +712,7 @@ fn php_module_internal(args: TokenStream2, input: TokenStream2) -> TokenStream2
 /// - `#[php(optional = i)]` - Sets the first optional parameter. Note that this
 ///   also sets the remaining parameters as optional, so all optional parameters
 ///   must be a variant of `Option<T>`.
-/// - `#[php(public)]`, `#[php(protected)]` and `#[php(private)]` - Sets the
+/// - `#[php(vis = "public")]`, `#[php(vis = "protected")]` and `#[php(vis = "private")]` - Sets the
```
Xenira: Don't update these docs manually please.

Change it in the guides and use the script in `tools` to update these docs.

src/builders/class.rs
```diff
@@ -238,16 +238,18 @@ impl ClassBuilder {
...
+
+        let mut func = FunctionBuilder::new("__construct", constructor::<T>);
+        let mut Visibility = MethodFlags::Public;
+        if let Some(ConstructorMeta {
+            build_fn, flags, ..
+        }) = T::constructor()
+        {
+            func = build_fn(func);
+            Visibility = flags.unwrap_or(MethodFlags::Public);
+        }
```
Xenira:
```suggestion
        let (func, visibility) = if let Some(ConstructorMeta {
            build_fn, flags, ..
        }) = T::constructor()
        {
            (build_fn(func), flags.unwrap_or(MethodFlags::Public))
        } else {
          (FunctionBuilder::new("__construct", constructor::<T>), MethodFlags::Public)
        }
```

src/builders/class.rs
```diff
@@ -238,16 +238,18 @@ impl ClassBuilder {
...
+
+        self.method(func, Visibility)
```
Xenira:
```suggestion
        self.method(func, visibility)
```

Xenira (changes requested):
Thanks, looks good. Just some minor code style suggestions.

src/class.rs
```diff
@@ -82,6 +82,8 @@ pub struct ConstructorMeta<T> {
     /// Function called to build the constructor function. Usually adds
     /// arguments.
     pub build_fn: fn(FunctionBuilder) -> FunctionBuilder,
+    /// Add constructor modification
+    pub flags: Option<MethodFlags>,
```
Xenira, line 86: Prob gonna move that to function builder later, but ok here for now I guess.

Norbytus: I think about this, but it could take more time and code changes. I just realized this problem, while implementing interface. Cause I always get error, in php, about __construct. I start lookin at this problem. I start from smallest.
But in future there should be opportunity not declare __construct.

For example interface could have __construct definition or not, and it's always abstract.
But abstract class can have abstract __construct and real defection, or doesn't have __construct. But I think abstract class its feature problem)))

I think in future there is a lot of work in macros. I agree that should be some more pretty variant to declare function/methods.

But if you have some recommendation or idea, I can check and try did better))

Xenira: Yeah, for now this is fine I guess. Will check if we can do this in a better way without bloating this MR too much.
If not I am fine with this and would move this at a later point in time.

crates/macros/src/lib.rs
Norbytus: I need help with this script, i try to run 
```
./tools/update_lib_docs.sh
```
And its change to old not working variant.
So i definitely doing something wrong

Xenira: You need to update the guides I linked

Norbytus: Done

Xenira (approved):
Sorry for the delay. Looks good and thank you ❤️

Norbytus:
> Sorry for the delay. Looks good and thank you ❤️

No problem

PR #561, opened 2025-10-19
feat: Add musl support

ptondereau:
## Description

- We're adding a Docker build specific to musl, as the embedded PHP lib also needs to be compiled for musl, and the official images only have embedded SAPI for ZTS Alpine.
- Rust doctests are disabled because `rustdoc --test` has its own compilation pipeline that overrides some flags with `-static-pie`
- We're also re-exporting the `runtime` and `static` features for the whole project.
- `embed` tests are disabled for now.

Note: The Dockerfile could be improved by a lot. I did my job here, as I'm not a Docker expert. 😅

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,14 +8,20 @@ apt update -y
 apt install -y \
   libclang-dev \
   bison \
-  re2c
+  re2c \
+  curl \
+  jq
+
+# Download and extract PHP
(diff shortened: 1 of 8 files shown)


ptondereau:
waiting #566 to be merged first

Dockerfile
Xenira, line 1: What is the advantage of downloading the release file over uaing git. Seems more complicated for the same result.

ptondereau: git is basically slower at downloading even if it's backed by Microsoft. 
You also skip ./buildconf

PR #565, opened 2025-10-28
chore: Add ptondereau to maintainer list

ptondereau:


Xenira:
Thanks for being so active. Hope to get back into it more, but too much other stuff coming up.

ptondereau:
> Thanks for being so active. Hope to get back into it more, but too much other stuff coming up.

No worries, actually we should focus on migrating into the php-rs org now because I have maybe a company that could sponsor the project.
We can sync up for that if you'd like.

PR #566, opened 2025-10-28
chore(rust): bump Rust edition to 2024

ptondereau:
## Breaking changes

Mostly `#[unsafe(no_mangle)]`

## Addition
- Added more tests for variadic arguments for array conversion


crates/macros/src/function.rs
```diff
@@ -218,6 +218,19 @@ impl<'a> Function<'a> {
...
+                let name = arg.name;
+                let variadic_name = format_ident!("__variadic_{}", name);
+                let clean_ty = arg.clean_ty();
+                Some(quote! {
+                    let #variadic_name = #name.variadic_vals::<#clean_ty>();
+                })
+            } else {
+                None
+            }
+        });
```
Xenira, line 232: Is this part of this mr? Seems like something unrelated.

tests/src/integration/variadic_args/mod.rs
Xenira, line 1: Same, maybe extract into separate mr

crates/macros/src/function.rs
ptondereau: this is because in 2024 edition,

```rust
#[php_function]
pub fn test_variadic(params: &[&Zval]) -> Vec<Zval> {
    params.iter().map(|x| x.shallow_clone()).collect()
}
```
fails to compile because Rust tells us that the temporary Vec for params is dropped too early.
I just store the args manually

tests/src/integration/variadic_args/mod.rs
ptondereau: this is because, I need to make sure that this part doesn't break while adding the previous patch related

crates/macros/src/function.rs
Xenira: Ahh ok, seems fine then.

Xenira (approved):

ptondereau:
tested against my project https://github.com/ptondereau/biscuit-php and it build fine

PR #576, opened 2025-10-29
feat(builders): add send_headers setter to SapiBuilder

Qard:
## Description

I need `send_headers` (plural) rather than just `send_header` (singular) to be able to know when sending of _all_ headers completes. I'm using this to support streaming so I can return a response body stream as soon as headers are available and then I can read from it while the request continues to be processed in another thread.

## Checklist


ptondereau:
Seems good enough, thanks!

PR #600, opened 2025-12-10
fix(ci): disable TS build for embed tests

ptondereau:
## Description

ZTS `libphp-embed` is not available (see: https://codeberg.org/oerdnj/deb.sury.org/issues/49) and building a whole PHP "just" for this case is too long for now.
Let's disable it


Xenira:
We could probably have a prebuild docker image to speed this up. Not a high priority, but ill create an issue for that.

Issue #508, opened 2025-07-13
A binary-safe way to write to PHP's stdout/stderr


kakserpom:
[php_print!](https://docs.rs/ext-php-rs/0.14.1/ext_php_rs/macro.php_print.html) panics on NUL.

I am looking for an idiomatic way to write to these virtual pipes.

I sure could use `Function::try_from_function("fwrite")` + try_call() + STDOUT/STDERR constants, there has to be another way :-)

ptondereau:
maybe a `php_error(&ErrorType::UserNotice, "Debug information");` could do the trick while providing a safer implementation of `php_print!`

kakserpom:
@ptondereau I am working with process pipes, so it's not really an option, but thanks.

Xenira:
From what I can see the only option would be to use `php_write` as the other functions take a c string.

https://github.com/php/php-src/blob/master/main/main.c#L943-L981

PR #613, opened 2025-12-18
feat(php_write): A binary-safe way to write to PHP's stdout/stderr #508

kakserpom:
Implements #508

### DIFF
--- a/guide/src/output.md
+++ b/guide/src/output.md
@@ -1,0 +1,136 @@
+# Output
+
+`ext-php-rs` provides several macros and functions for writing output to PHP's
+stdout and stderr streams. These are essential when your extension needs to
+produce output that integrates with PHP's output buffering system.
+
+## Text Output
+
+For regular text output (strings without NUL bytes), use the `php_print!` and
+`php_println!` macros. These work similarly to Rust's `print!` and `println!`
+macros.
+
+### `php_print!`
+
+Prints to PHP's standard output without a trailing newline.
+
+```rust,ignore
+use ext_php_rs::prelude::*;
+
+#[php_function]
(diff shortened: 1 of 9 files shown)


src/zend/mod.rs
```diff
@@ -62,6 +63,81 @@ pub fn printf(message: &str) -> Result<()> {
...
+    use std::io::Write;
+
+    if data.is_empty() {
+        return 0;
+    }
+
+    std::io::stderr()
+        .write(data)
+        .unwrap_or(0)
+}
```
ptondereau, line 141: I don't see any benefit to having this wrapper in our lib.
Maybe a small doc is enough

src/zend/mod.rs
```diff
@@ -62,6 +63,43 @@ pub fn printf(message: &str) -> Result<()> {
...
+#[must_use]
+pub fn write(data: &[u8]) -> usize {
+    unsafe {
+        if let Some(ub_write) = sapi_module.ub_write {
+            ub_write(data.as_ptr().cast::<c_char>(), data.len())
+        } else {
+            0
+        }
+    }
+}
```
ptondereau: We should use a `Result<usize>` instead for better error handling instead of just returning 0.

guide/src/output.md
```diff
@@ -0,0 +1,112 @@
+# Output
```
ptondereau, line 1: We should add a note with possible output interleaving on ZTS builds and show atomic output with buffering like:

```php
<?php

ob_start();
rust_function(); // with some php_writeln!
$output = ob_get_clean();
echo $output;
```

It remains thread-safe anyway, but just to show all info for the end user

PR #615, opened 2025-12-18
fix(macro): reference mutability inside Option #515

### LINKED ISSUE #515: Argument #1 ($foo) could not be passed by reference
```
use ext_php_rs::types::ZendHashTable;
use ext_php_rs::prelude::*;

#[php_class]
pub struct Example {}

#[php_impl]
impl Example {
    fn __construct() -> Self {
        Self {}
    }
    pub fn foo(&self, foo: Option<&ZendHashTable>) {}
}

#[php_module]
    module.class::<Example>()
}
```

```
<?php
$example = new Example;
$example->foo([1,2,3]);
echo $bar;
```

Note that it works without `Option<...>`.

This works too:
```
$foo = [1,2,3];
$example->foo($foo);
```

kakserpom:
Fixes #515

### DIFF
--- a/src/types/zval.rs
+++ b/src/types/zval.rs
@@ -14,8 +14,8 @@ use crate::{
     error::{Error, Result},
     ffi::{
         _zval_struct__bindgen_ty_1, _zval_struct__bindgen_ty_2, ext_php_rs_zend_string_release,
-        zend_is_callable, zend_is_identical, zend_is_iterable, zend_resource, zend_value, zval,
-        zval_ptr_dtor,
+        zend_array_dup, zend_is_callable, zend_is_identical, zend_is_iterable, zend_resource,
+        zend_value, zval, zval_ptr_dtor,
     },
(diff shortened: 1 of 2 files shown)


ptondereau:
~bugfix(macro): reference mutability inside Option~
fix(macro): reference mutability inside Option

tests/src/integration/array/mod.rs
```diff
@@ -41,13 +41,20 @@ pub fn test_array_keys() -> Zval {
     ht.into_zval(false).unwrap()
 }
 
+/// Test that `Option<&ZendHashTable>` can accept literal arrays (issue #515)
+#[php_function]
+pub fn test_optional_array_ref(arr: Option<&ZendHashTable>) -> i64 {
```
ptondereau, line 46: can we also add a test with `arr: Option<&mut ZendHashTable>`? It would add an anti regression test for mutable param

PR #616, opened 2025-12-18
fix(macro): identifier-related bugs #536

### LINKED ISSUE #536: macros: raw identifier bug
Branch: master

```
use ext_php_rs::prelude::*;

#[php_class]
pub struct Example {}
#[php_impl]
impl Example {
    pub fn foo(&self, r#as: &str) {}
}
#[php_module]
    module.class::<Example>()
}
```

```
<?php

// Stubs for untitled5

namespace {
    class Example {
        public function foo(string $r#as) {}

        public function __construct() {}
    }
}
```

Also, `__construct()` in stubs is unnecessary since you can't instantiate this class from PHP.

P.S. Pardon me for not using the issue templates :-)

kakserpom:
Fixes #536

### DIFF
--- a/crates/macros/src/enum_.rs
+++ b/crates/macros/src/enum_.rs
@@ -8,7 +8,9 @@ use syn::{Fields, Ident, ItemEnum, Lit};
 
 use crate::{
     helpers::get_docs,
-    parsing::{PhpRename, RenameRule, Visibility},
+    parsing::{
+        PhpNameContext, PhpRename, RenameRule, Visibility, ident_to_php_name, validate_php_name,
+    },
     prelude::*,
 };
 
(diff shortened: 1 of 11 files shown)


crates/macros/src/parsing.rs
```diff
@@ -1,6 +1,213 @@
...
+    "self",
+    "parent",
+];
+
+/// Type keywords that are reserved for class/interface/enum names but CAN be
+/// used as method, function, constant, or property names in PHP.
+const PHP_TYPE_KEYWORDS: &[&str] = &[
+    "bool", "false", "float", "int", "iterable", "mixed", "never", "null", "numeric", "object",
+    "resource", "string", "true", "void",
+];
```
ptondereau, line 126: Looking at https://github.com/php/php-src/blob/0880433e61d3c5c350f73f149ed1086200712177/Zend/zend_compile.c#L202

`resource` and `numeric` are allowed for class name

crates/macros/src/enum_.rs
```diff
@@ -84,7 +86,7 @@ pub fn parser(mut input: ItemEnum) -> Result<TokenStream> {
         cases.push(EnumCase {
             ident: variant.ident.clone(),
```
ptondereau, line 93: I think we should also validate this one

crates/macros/src/enum_.rs
```diff
@@ -84,7 +86,7 @@ pub fn parser(mut input: ItemEnum) -> Result<TokenStream> {
...
             name: variant_attr.rename.rename(
-                variant.ident.to_string(),
+                ident_to_php_name(&variant.ident),
                 php_attr.rename_cases.unwrap_or(RenameRule::Pascal),
```
ptondereau: And also validate cases (eg: `case True` doesn't work)

crates/macros/src/parsing.rs
```diff
@@ -1,6 +1,213 @@
...
+    };
+
+    assert!(
+        !is_forbidden,
+        "Cannot use '{}' as a PHP {} name: '{}' is a reserved keyword in PHP. \
+         Consider using a different name or the #[php(name = \"...\")] attribute to specify an alternative PHP name.",
+        name,
+        context.description(),
+        name
+    );
```
ptondereau: we should use syn:Error handling for macro usage

kakserpom:
@ptondereau please take a look again

ptondereau (approved):

PR #617, opened 2025-12-18
fix(macro): nullable parameters #538

### LINKED ISSUE #186: [stubs] class constants values are always null
| Component | Version |
| :---              |     :---:    | 
| ext_php_rs  | git:master |
| cargo-php  |  v0.1.7 |

Extension code
```rust
#[php_class]
pub struct Foo;
#[php_impl]
impl Foo {
  const BAR: &'static str = "baz";
  const BAZ: i32 = 42;
}
```
Expected stubs:
```php
<?php

class Foo {
  public const BAR = "baz";
  public const BAZ = 42;
}
```
Actual stubs:
```php
<?php

class Foo {
  const BAR = null; 
  const BAZ = null;
}
```

### LINKED ISSUE #305: Stub generation of optional arguments doesn't set default value to NULL
Given a function like this:

```rust
#[php_function()]
pub fn foo(a: String, b: Option<bool>) -> PhpResult<Zval> {
    ...
}
```

The generated stub looks like this:
```php
namespace {
    function foo(string $a, ?bool $b): mixed {}
}
```

However, this is not reflecting the truth, as `$b` here is just documented as nullable, but **must** be specified, whereas in reality it's optional. The stub should therefore be

```php
namespace {
    function foo(string $a, ?bool $b = null): mixed {}
}
```

### LINKED ISSUE #538: Generate nullable param in function signature
## Description
When declare function/method with macros, it's generate nullable params. 

## Steps to Reproduce
In rust declare a struct with function with default value. In macro crate function.rs on nullable check it's set nullable for default value.
For me it's weird, but maybe only for me)

3. 

## Example

**Extension Code:**
```rs
#[php_class]
#[php(name = "Rust\\A")]
struct A;

#[php_impl]
impl A {
    #[php(defaults(value = 0))]
    pub fn test(value: i32) -> i32 {
        value
    }
}
```

**PHP Code:**
```php
<?php
$test = new Rust\A;
// Will work find
var_dump($test->test(null));

```

### Expected Behavior
```
Should be error.
```

### Log Output

```shell

```

### Affected Components

- [ ] ext-php-rs
- [x] macros
- [ ] cli (cargo-php)

### PHP Version

```plain
PHP 8.4
```

### `ext-php-rs` Version

0.4.12

### Operating System

Archlinux

kakserpom:
Fixes #538 
Fixes #186
Fixes #305

### DIFF
--- a/crates/macros/src/function.rs
+++ b/crates/macros/src/function.rs
@@ -637,7 +637,7 @@ impl TypedArg<'_> {
             None
         };
         let default = self.default.as_ref().map(|val| {
-            let val = val.to_token_stream().to_string();
+            let val = expr_to_php_stub(val);
             quote! {
                 .default(#val)
             }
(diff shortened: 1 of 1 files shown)


ptondereau:
`bugfix` doesn't exist in conventional commit.

kakserpom:
@ptondereau What a weird CI fail...

tests/src/integration/defaults/defaults.php
```diff
@@ -7,3 +7,16 @@
...
+$threw = false;
+try {
+    test_defaults_integer(null);
+} catch (TypeError $e) {
+    $threw = true;
+}
+assert($threw, 'Expected TypeError when passing null to non-nullable parameter with default');
+
+// But passing null to a nullable parameter should still work
+assert(test_defaults_nullable_string(null) === null);
```
ptondereau, line 22: can we also add a test for something like:
```rust
#[php_function]
#[php(defaults(a = Some("fallback".to_string())))]
pub fn test_defaults_nullable_with_some_default(a: Option<String>) -> Option<String> {
    a
}
```
```php
assert(test_defaults_nullable_with_some_default() === 'fallback');
assert(test_defaults_nullable_with_some_default(null) === null);
assert(test_defaults_nullable_with_some_default('custom') === 'custom');
```

PR #620, opened 2025-12-18
fix(bindings): exclude `preserve_none` instruction from bindings

ptondereau:
## Description

`bindgen` doesn't support calling the `preserve_none` convention that has been introduced in the last PHP 8.5.1.
We can't exclude it with Clang Builder nor passing CFLAGS because this convention is used everywhere in the PHP src code.

I've forked bindgen + clang-sys to be able to build an extension on arm64 with macOS, but we clearly need an update from these repositories.

PHP issue: https://github.com/php/php-src/issues/20546

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,8 +79,10 @@ path = "tests/module.rs"
 name = "sapi_tests"
 path = "tests/sapi.rs"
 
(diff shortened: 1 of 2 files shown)


ptondereau:
@Xenira if you have a better move...

PR #622, opened 2025-12-20
fix(zval): Heap corruption with persistent=true #424

### LINKED ISSUE #424: Core dumped where try Zval::set_string with persist trye
```rust
use ext_php_rs::prelude::*;
use ext_php_rs::types::Zval;

pub fn startup(_ty: i32, _: i32) -> i32 {
    0
}

#[php_class]
#[derive(Default)]
struct Test {}

#[php_impl]
impl Test {
    #[php(constructor)]
    fn __constructor() -> Self {
        Self {}
    }
    fn test(&self) {
        let mut z = Zval::new();
        let _ = z.set_string("TEST", true);
    }
}

#[php_module]
fn module(module: ModuleBuilder) -> ModuleBuilder {
    module.class::<Test>()
}
```

```php
<?php

$t = new Test;

var_dump($t->test());
```

Output
```bash
zend_mm_heap corrupted
[2]    41472 IOT instruction (core dumped)  php -d extension=./target/debug/libtest_php_ext.so ./index.php
```

OS: Archlinux

kakserpom:
Fixes #424

### DIFF
--- a/tests/src/integration/persistent_string/mod.rs
+++ b/tests/src/integration/persistent_string/mod.rs
@@ -1,0 +1,69 @@
+use ext_php_rs::prelude::*;
+use ext_php_rs::types::Zval;
+
+#[php_function]
+pub fn test_persistent_string() -> String {
+    let mut z = Zval::new();
+    let _ = z.set_string("PERSISTENT_STRING", true);
+    // z is dropped here - this was causing heap corruption before the fix
+    "persistent string test passed".to_string()
+}
+
+#[php_function]
+pub fn test_non_persistent_string() -> String {
+    let mut z = Zval::new();
+    let _ = z.set_string("NON_PERSISTENT_STRING", false);
+    "non-persistent string test passed".to_string()
+}
+
+#[php_function]
+pub fn test_persistent_string_read() -> String {
+    let mut z = Zval::new();
+    let _ = z.set_string("READ_BEFORE_DROP", true);
+    let value = z.str().unwrap_or("failed");
+    format!("read: {value}")
+}
+
+#[php_function]
+pub fn test_persistent_string_loop(count: i64) -> String {
+    for i in 0..count {
+        let mut z = Zval::new();
+        let s = format!("LOOP_{i}");
+        let _ = z.set_string(&s, true);
+    }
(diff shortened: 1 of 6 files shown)


tests/src/integration/persistent_string/mod.rs
```diff
@@ -0,0 +1,79 @@
+//! Integration tests for issue #424 - `Zval::set_string` with persistent flag
+//!
+//! Tests that persistent strings are correctly handled when a Zval is dropped.
```
ptondereau:
```suggestion
```

I think this useless

tests/src/integration/persistent_string/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use ext_php_rs::prelude::*;
+use ext_php_rs::types::Zval;
+
+/// Test persistent string creation and drop
```
ptondereau: All these comments before the test function are useless because the function name should reflect the test case.

tests/src/integration/persistent_string/persistent_string.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+/**
+ * Integration test for issue #424 - Zval::set_string with persistent flag
+ *
+ * Tests that persistent strings are correctly handled when a Zval is dropped,
+ * without causing heap corruption.
+ */
```
ptondereau: same here, we know what we're doing by the name of the file

tests/src/integration/persistent_string/persistent_string.php
```diff
@@ -0,0 +1,34 @@
...
+
+// Test 5: Interned string (persistent)
+$result = test_interned_string_persistent();
+assert($result === "interned persistent test passed", "Interned persistent string test failed: $result");
+
+// Test 6: Interned string (non-persistent)
+$result = test_interned_string_non_persistent();
+assert($result === "interned non-persistent test passed", "Interned non-persistent string test failed: $result");
+
+echo "All persistent string tests passed!\n";
```
ptondereau: Useless as all assertions are the guards

kakserpom:
@ptondereau Done

Issue #310, opened 2024-01-24
Slower Than C 


arshidkv12:
Hi
I tested with xdebug profile option with grind cache. 
Functions are slower than c. How to speed up?

ryangjchandler:
The difference in speed is likely due to converting between PHP types (`Zval`) and Rust types such as `String`, `Vec`, etc.

joehoyle:
I think to test this assumption it could be a good idea to benchmark:

```
#[php_function]
fn should_be_fast(&Zval zval ) -> Zval {
    zval.shallow_clone()
}
```

I suspect we might actually have several inefficiencies in the calling bridging, unnecessarily allocating etc.

ryangjchandler:
That would just be benchmarking a direct clone of a Zval though right? Not the `Vec<T>` allocations, etc?

joehoyle:
I'm not sure, I think the `#[php_function]` may generate code that has additional allocations etc in.

EdmondDantes:
The current approach is expected to introduce some performance overhead:

Calling a function from a wrapper function.
Copying variables to the stack.
In other words, if I create a function whose body does very little useful work and plan to call this function frequently, I will inevitably lose about 50% of performance.

I'm not sure how common this case is, but if necessary, it's possible to create a function without a wrapper and work directly with zend_execute_data* execute_data and zval* return_val.

Xenira:
For calling functions more efficiently i use something like this: https://github.com/Xenira/php-iter-tools/blob/rust-impl/src/lib.rs#L1739

arshidkv12:
I think zig is best for it https://github.com/arshidkv12/zig-php

EdmondDantes:
@arshidkv12 

```zig
export fn hello_world(execute_data: ?*php.zend_execute_data, return_value: ?*php.zval) void {
    _ = execute_data;
    _ = return_value;
    _ = php.php_printf("Hello from ZIG!\n");
}
```

@Xenira 

That's exactly what I'm talking about – the same code can be written in Rust without any performance loss. We need to consider this implementation option.

kakserpom:
@Xenira
Why don't we move [your implementation](https://github.com/Xenira/php-iter-tools/blob/rust-impl/src/lib.rs#L1739) from php-iter-tools into ext-php-rs?

Xenira:
@kakserpom is on my todo list. Will create an issue for this.

Xenira:
Probably want to create performance tests first though

[...]

PR #602, opened 2025-12-11
test(benches): add basic benchmark infrastructure

Xenira:
Refs: #599

## Description

Adds basic benchmarking infrastructure

TODO:

- [x] CI/CD integration
- [x] Some more basic benchmarks as baseline

## Checklist

_Check the boxes that apply (put an `x` in the brackets, like `[x]`). You can also check boxes after the PR is created._

- [ ] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [ ] I have added tests that prove my code works as expected.
- [ ] I have added documentation if applicable.
- [ ] I have added a [migration guide](../CONTRIBUTING.md#breaking-changes) if applicable.

:heart: Thank you for your contribution!

PR #623, opened 2025-12-20
feat(functions): CachedCallable and raw functions #310

kakserpom:
Implements #310


src/types/callable.rs
```diff
@@ -190,3 +193,441 @@ impl Deref for OwnedZval<'_> {
...
+///
+/// let mut callback = CachedCallable::try_from_name("strtoupper").unwrap();
+///
+/// // Each call reuses the cached function info
+/// for word in ["hello", "world", "rust"] {
+///     let result = callback.try_call(vec![&word]).unwrap();
+///     println!("Upper: {}", result.string().unwrap());
+/// }
+/// ```
+pub struct CachedCallable {
```
ptondereau, line 296: We need a implementation for `Drop` that dtor `self.fci.function_name`

kakserpom: No need for that, `zval_ptr_dtor` is getting called already.

Xenira (changes requested):
Should use platform independent benchmarks from #602 for this

kakserpom:
> Should use platform independent benchmarks from #602 for this

Well #602 is still in the works, you can merge this one and then replace the benchmarks with yours in #602 or you can merge #602 first...

Xenira:
@kakserpom I will try to have all benchmarks in #602 to get a baseline.

Another thing we might want to consider is having the plain callable initialize the function info cach lazily to have it perform better as well. Not sure if we should just merge those two and have the option to set it to lazy / eager / no caching.

kakserpom:
@Xenira 

> Another thing we might want to consider is having the plain callable initialize the function info cach lazily to have it perform better as well. Not sure if we should just merge those two and have the option to set it to lazy / eager / no caching.

Done.

src/types/callable.rs
```diff
@@ -190,3 +265,422 @@ impl Deref for OwnedZval<'_> {
...
+    /// The callable zval (keeps the callable alive)
+    callable: Zval,
+    /// Cached function call info
+    fci: zend_fcall_info,
```
ptondereau: `fci` should not be stored in the struct:

we mutate it later in the try_call, and as we re-invoke the same callable (re-entrancy), I think this is UB.

If you take a look at `_call_user_function_impl` from PHP source code, `fci` is always stack-local, and that's why `ZendCallable` is safe now.

Workaround can be to construct `zend_fcall_info` on the stack for each `try_call`

src/types/callable.rs
```diff
@@ -190,3 +265,422 @@ impl Deref for OwnedZval<'_> {
...
+
+        // Convert parameters to zvals
+        let params: Vec<Zval> = params
+            .into_iter()
+            .map(|val| val.as_zval(false))
+            .collect::<Result<Vec<_>>>()?;
+
+        // Update fci with the current call parameters
+        self.fci.retval = &raw mut retval;
+        self.fci.params = params.as_ptr().cast_mut();
```
ptondereau: this is here that could cause UB

Issue #325, opened 2024-10-10
Returning static str reference from class getter fail to compile


dmarteau:
Returning static str reference from class getter fail to build witth the error: `
implementation of `FromZval` is not general enough`

### Example:

```
use ext_php_rs::prelude::*;

#[php_class]
struct Test;

#[php_impl]
impl Test {

    #[getter]
    fn get_static_value(&self) -> &'static str {
        "Hello"
    }
}

#[php_module]
pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
    module
}
```

Fail to complie with the error:

```
error: implementation of `FromZval` is not general enough
   --> proj4rs-php/src/lib.rs:188:1
    |
188 | #[php_module]
    | ^^^^^^^^^^^^^ implementation of `FromZval` is not general enough
    |
    = note: `FromZval<'0>` would have to be implemented for the type `&str`, for any lifetime `'0`...
    = note: ...but `FromZval<'1>` is actually implemented for the type `&'1 str`, for some specific lifetime `'1`
    = note: this error originates in the attribute macro `php_module` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Removing the `#[getter]` macro makes compilation succeed.

dmarteau:
platform:  Ubuntu 24.04
php :  8.3.6 
rustc: 1.81.0

kakserpom:
@ptondereau Let's close this?

PR #624, opened 2025-12-20
feat(class): Getter/setter implementation #325

kakserpom:
Implements #325

### DIFF
--- a/crates/macros/src/class.rs
+++ b/crates/macros/src/class.rs
@@ -312,9 +312,21 @@ fn generate_registered_class_impl(
                 &'static str, ::ext_php_rs::internal::property::PropertyInfo<'a, Self>
             > {
                 use ::std::iter::FromIterator;
-                ::std::collections::HashMap::from_iter([
+                use ::ext_php_rs::internal::class::PhpClassImpl;
+
+                // Start with field properties (instance fields only, not static)
+                let mut props = ::std::collections::HashMap::from_iter([
                     #(#instance_fields,)*
-                ])
+                ]);
+
+                // Add method properties (from #[php(getter)] and #[php(setter)])
+                let method_props = ::ext_php_rs::internal::class::PhpClassImplCollector::<Self>::default()
+                    .get_method_props();
+                for (name, prop_info) in method_props {
+                    props.insert(name, prop_info);
+                }
+
+                props
             }
 
             #[must_use]
--- a/crates/macros/src/impl_.rs
+++ b/crates/macros/src/impl_.rs
@@ -163,6 +163,21 @@ impl MethodArgs {
     }
 }
 
+/// A property getter or setter method.
+#[derive(Debug)]
+struct PropertyMethod<'a> {
+    /// Property name in PHP (e.g., "name" for `get_name`/`set_name`).
+    prop_name: String,
+    /// The Rust method identifier.
+    method_ident: &'a syn::Ident,
+    /// Whether this is a getter (true) or setter (false).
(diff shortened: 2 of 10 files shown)


ptondereau:
@kakserpom could you rebase please

kakserpom:
@ptondereau done

kakserpom:
@ptondereau Just rebased again :)

crates/macros/src/impl_.rs
```diff
@@ -262,6 +276,32 @@ impl<'a> ParsedImpl<'a> {
                     method.attrs.retain(|attr| !attr.path().is_ident("php"));
 
                     let opts = MethodArgs::new(name, attr)?;
+
+                    // Handle getter/setter methods
+                    if matches!(opts.ty, MethodTy::Getter | MethodTy::Setter) {
+                        let is_getter = matches!(opts.ty, MethodTy::Getter);
+                        // Extract property name by stripping get_/set_ prefix
+                        let method_name = method.sig.ident.to_string();
+                        let prop_name = if is_getter {
```
ptondereau, line 291: opts.name is computed but never used, so users cannot rename getter/setter-backed properties. Is this expected?

crates/macros/src/class.rs
```diff
@@ -312,9 +312,25 @@ fn generate_registered_class_impl(
...
+                ]);
+
+                // Add method properties (from #[php(getter)] and #[php(setter)])
+                let method_props = ::ext_php_rs::internal::class::PhpClassImplCollector::<Self>::default()
+                    .get_method_props();
+                for (name, prop) in method_props {
+                    props.insert(name, ::ext_php_rs::internal::property::PropertyInfo {
+                        prop,
+                        flags: ::ext_php_rs::flags::PropertyFlags::Public,
+                        docs: &[],
```
ptondereau: this is sad that we lose the doc comment :'(

crates/macros/src/class.rs
```diff
@@ -312,9 +312,25 @@ fn generate_registered_class_impl(
...
-                ])
+                ]);
+
+                // Add method properties (from #[php(getter)] and #[php(setter)])
+                let method_props = ::ext_php_rs::internal::class::PhpClassImplCollector::<Self>::default()
+                    .get_method_props();
+                for (name, prop) in method_props {
+                    props.insert(name, ::ext_php_rs::internal::property::PropertyInfo {
+                        prop,
+                        flags: ::ext_php_rs::flags::PropertyFlags::Public,
```
ptondereau: we ignore the user prop `#[php(vis = "...")]` here :/

crates/macros/src/class.rs
```diff
@@ -312,9 +312,25 @@ fn generate_registered_class_impl(
...
                 use ::std::iter::FromIterator;
-                ::std::collections::HashMap::from_iter([
+                use ::ext_php_rs::internal::class::PhpClassImpl;
+
+                // Start with field properties (instance fields only, not static)
+                let mut props = ::std::collections::HashMap::from_iter([
                     #(#instance_fields,)*
-                ])
+                ]);
+
```
ptondereau, line 321: I think we should document how to use a pub prob and use it in a getter implementation.

```Rust
#[php_class]
pub struct Book {
  #[php(prop)]
  pub title: String,
}

#[php_impl]
impl Book {
  #[php(getter)]
  pub fn get_title(&self) -> String { // now $book->title calls this getter instead of the above prop
  }
}
```

ptondereau (approved):

PR #625, opened 2025-12-20
fix(zend_bailout): Fix zend_bailout handling #537

### LINKED ISSUE #537: PHP bypass Rust destructors when `zend_bailout`
### Description
From discord discussion:

> _emalloc will call zend_bailout should PHP's memory limit be reached. In the background, this will do a long jump bypassing any Rust destructors (effectively not freeing memory).

> We would simply end up intercepting the bailout (like here: https://gist.github.com/awons/d87db73d1ad5eae1445b9de7e3727c6a) and then panicking. But that panic needs to be caught before it crosses the FFI boundary. And then, in C we would need to again bail to make sure the OOM works as it should.

Long jump located [here](https://github.com/php/php-src/blob/d20f4fca69ce82830f5ca2b8cdf2aa26c4c89909/Zend/zend.c#L1268)

## Steps to Reproduce
https://github.com/ptondereau/php-rust-bailout-repro

### Affected Components

- [x] ext-php-rs
- [ ] macros
- [ ] cli (cargo-php)

### PHP Version

```plain
Every PHP version
```

### `ext-php-rs` Version

every

### Operating System

every

kakserpom:
Fixes #537

### DIFF
--- a/src/embed/mod.rs
+++ b/src/embed/mod.rs
@@ -17,7 +17,7 @@ use crate::types::{ZendObject, Zval};
 use crate::zend::{ExecutorGlobals, panic_wrapper, try_catch};
 use parking_lot::{RwLock, const_rwlock};
 use std::ffi::{CString, NulError, c_char, c_void};
-use std::panic::{RefUnwindSafe, resume_unwind};
+use std::panic::{AssertUnwindSafe, UnwindSafe, resume_unwind};
 use std::path::Path;
 use std::ptr::null_mut;
 
@@ -105,7 +105,9 @@ impl Embed {
             zend_stream_init_filename(&raw mut file_handle, path.as_ptr());
         }
 
-        let exec_result = try_catch(|| unsafe { php_execute_script(&raw mut file_handle) });
+        let exec_result = try_catch(AssertUnwindSafe(|| unsafe {
+            php_execute_script(&raw mut file_handle)
+        }));
 
         unsafe { zend_destroy_file_handle(&raw mut file_handle) }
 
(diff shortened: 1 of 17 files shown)


ptondereau:
@kakserpom can you rebase, please? This is the next PR I want to merge.

src/embed/mod.rs
```diff
@@ -161,6 +163,9 @@ impl Embed {
             )
         };
 
+        // Prevent the closure from being dropped here since it was consumed in panic_wrapper
+        std::mem::forget(func);
+
```
ptondereau, line 168: I think this is not true that we prevent anything here: `func` is already consumed in `let func = unsafe { std::ptr::read(ctx.cast::<F>()) }; catch_unwind(func);`. But don't trust me on this one, and test it.

src/zend/try_catch.rs
```diff
@@ -76,6 +79,9 @@ fn do_try_catch<R, F: FnMut() -> R + RefUnwindSafe>(func: F, first: bool) -> Res
         }
     };
 
+    // Prevent the closure from being dropped here since it was consumed in panic_wrapper
+    std::mem::forget(func);
```
ptondereau, line 83: same here

ptondereau (dismissed):

src/embed/mod.rs
kakserpom: I've tested it and it's correct as is.

ptondereau: This is weird, as `std::ptr::read` moves the closure out of memory.

kakserpom: forget() is needed to avoid a double-drop.

kakserpom:
@ptondereau A spurious CI failure...

src/embed/mod.rs
ptondereau: ok TIL

PR #626, opened 2025-12-20
fix(class): Return Self ($this) #502

### LINKED ISSUE #502: Returning $this
```
#[php_class]
struct Foo {}
#[php_impl]
impl Foo {
    pub fn someMethod(&mut self) -> &mut Self {
        self
    }
}
```

```
error[E0277]: the trait bound `&mut PhpClassImplCollector<Foo>: IntoZval` is not satisfied
 --> src/lib.rs:7:37
  |
7 |     pub fn someMethod(&mut self) -> &mut Self {
  |                                     ^^^^^^^^^ the trait `IntoZval` is not implemented for `&mut PhpClassImplCollector<Foo>`
  |
```

kakserpom:
Fixes #502

### DIFF
--- a/crates/macros/src/function.rs
+++ b/crates/macros/src/function.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 
 use darling::{FromAttributes, ToTokens};
 use proc_macro2::{Ident, Span, TokenStream};
-use quote::{format_ident, quote};
+use quote::{format_ident, quote, quote_spanned};
 use syn::spanned::Spanned as _;
 use syn::{Expr, FnArg, GenericArgument, ItemFn, PatType, PathArguments, Type, TypePath};
 
@@ -10,6 +10,37 @@ use crate::helpers::get_docs;
 use crate::parsing::{PhpRename, RenameRule, Visibility};
 use crate::prelude::*;
 use crate::syn_ext::DropLifetimes;
+
+/// Checks if the return type is a reference to Self (`&Self` or `&mut Self`).
+/// This is used to detect methods that return `$this` in PHP.
+fn returns_self_ref(output: Option<&Type>) -> bool {
+    let Some(ty) = output else {
+        return false;
+    };
+    if let Type::Reference(ref_) = ty
+        && let Type::Path(path) = &*ref_.elem
+        && path.path.segments.len() == 1
+        && let Some(segment) = path.path.segments.last()
+    {
+        return segment.ident == "Self";
+    }
+    false
+}
+
+/// Checks if the return type is `Self` (not a reference).
+/// This is used to detect methods that return a new instance of the same class.
+fn returns_self(output: Option<&Type>) -> bool {
+    let Some(ty) = output else {
+        return false;
+    };
+    if let Type::Path(path) = ty
+        && path.path.segments.len() == 1
+        && let Some(segment) = path.path.segments.last()
+    {
+        return segment.ident == "Self";
+    }
+    false
+}
 
 pub fn wrap(input: &syn::Path) -> Result<TokenStream> {
     let Some(func_name) = input.get_ident() else {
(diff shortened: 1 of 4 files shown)


crates/macros/src/function.rs
```diff
@@ -11,6 +11,21 @@ use crate::parsing::{PhpRename, RenameRule, Visibility};
...
 
+/// Checks if the return type is a reference to Self (`&Self` or `&mut Self`).
+/// This is used to detect methods that return `$this` in PHP.
+fn returns_self_ref(output: Option<&Type>) -> bool {
+    let Some(ty) = output else {
+        return false;
+    };
+    if let Type::Reference(ref_) = ty
+        && let Type::Path(path) = &*ref_.elem
+        && let Some(segment) = path.path.segments.last()
```
ptondereau, line 38:
```suggestion
    if let Type::Reference(ref_) = ty
        && let Type::Path(path) = &*ref_.elem
        && path.path.segments.len() == 1
        && let Some(segment) = path.path.segments.last()
```
To be tested, but I think it avoids matching something like `foo::bar::Self`

crates/macros/src/function.rs
```diff
@@ -199,24 +214,54 @@ impl<'a> Function<'a> {
             }
         };
 
+        // Check if this method returns &Self or &mut Self
+        // In that case, we need to return `this` (the ZendClassObject) directly
```
ptondereau, line 250: We should add a check to enhance macro UX:
When static methods are returning `&Self` or `&mut Self `, we should warn the user.

```rust
if returns_self_ref(self.output) {
      if let CallType::Method { receiver: MethodReceiver::Static, .. } = call_type {
          return quote_spanned! { self.output.span() =>
              compile_error!(
                  "Static methods cannot return &Self or &mut Self. \
                   Only instance methods can use fluent interface pattern returning $this."
              )
          };
      }
  }

// the rest ...
```

crates/macros/src/function.rs
```diff
@@ -306,15 +369,27 @@ impl<'a> Function<'a> {
...
-                    MethodReceiver::ZendClassObject => {
+                    (MethodReceiver::Class, true) => {
+                        quote! { let _ = this.#ident(#({#arg_accessors}),*); }
+                    }
+                    (MethodReceiver::Class, false) => {
+                        quote! { this.#ident(#({#arg_accessors}),*) }
+                    }
+                    (MethodReceiver::ZendClassObject, true) => {
+                        quote! { let _ = #class::#ident(this, #({#arg_accessors}),*); }
+                    }
```
ptondereau, line 439:
```suggestion
                    (MethodReceiver::ZendClassObject, true) => {
					      quote! {
					          {
					              let _ = this.inner().#ident(#({#arg_accessors}),*);
					          }
					      }
				  	}
```

Explicit scope will help here, for when the method returns ` &mut Self`: this creates a mutable borrow of `$this` that's immediately discarded, and I guess the Rust compiler considers that as implicit.

crates/macros/src/function.rs
```diff
@@ -229,9 +274,24 @@ impl<'a> Function<'a> {
...
             output.drop_lifetimes();
+
+            // If returning &Self or &mut Self from a method, use the class type
+            // for return type information since we return `this` (ZendClassObject)
+            if returns_self_ref(self.output)
+                && let Some(CallType::Method { class, .. }) = call_type
+            {
+                return quote! {
+                    .returns(
+                        <&mut ::ext_php_rs::types::ZendClassObject<#class> as ::ext_php_rs::convert::IntoZval>::TYPE,
```
ptondereau, line 320: Can you verify that `IntoZval::TYPE` for `&mut ZendClassObject<T>` correctly resolves to `DataType::Object(Some(T::CLASS_NAME))` and not a wrapper type name?

kakserpom: Yes

Issue #252, opened 2023-07-16
Static property and method 


arshidkv12:
How to add static property and method?

``` #[static_]
    pub fn hello( q: String) -> String {
        format!("hello {}", q)
    }

**Comment by joehoyle** [comment]
@arshidkv12 I believe just that should work for static methods. From the docs:

> Class methods can take a `&self` or `&mut self` parameter. They cannot take a
consuming `self` parameter. Static methods can omit this `self` parameter.

**Comment by arshidkv12** [comment]
I want code as follows.

**PHP**
``` 
class my class{

  static $a = 1;
  public static function hello(){

  }
}

PR #627, opened 2025-12-20
feat(class): Static properties and methods #252

kakserpom:
Implements #252

### DIFF
--- a/guide/src/macros/classes.md
+++ b/guide/src/macros/classes.md
@@ -24,12 +24,16 @@ PHP property. By default, the field will be accessible from PHP publicly with
 the same name as the field. Property types must implement `IntoZval` and
 `FromZval`.
 
-You can rename the property with options:
+You can customize properties with these options:
 
 - `name` - Allows you to rename the property, e.g.
-  `#[php(name = "new_name")]`
+  `#[php(prop, name = "new_name")]`
 - `change_case` - Allows you to rename the property using rename rules, e.g.
-  `#[php(change_case = PascalCase)]`
+  `#[php(prop, change_case = PascalCase)]`
+- `static` - Makes the property static (shared across all instances), e.g.
+  `#[php(prop, static)]`
+- `flags` - Sets property visibility flags, e.g.
+  `#[php(prop, flags = ext_php_rs::flags::PropertyFlags::Private)]`
 
 ## Restrictions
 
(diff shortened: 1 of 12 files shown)


src/zend/class.rs
```diff
@@ -132,6 +135,74 @@ impl ClassEntry {
...
+    /// use ext_php_rs::zend::ClassEntry;
+    ///
+    /// let ce = ClassEntry::try_find("MyClass").unwrap();
+    /// let value: Option<i64> = ce.get_static_property("counter");
+    /// ```
+    #[must_use]
+    pub fn get_static_property<'a, T: FromZval<'a>>(&'a self, name: &str) -> Option<T> {
+        let name = CString::new(name).ok()?;
+        let zval = unsafe {
+            crate::ffi::zend_read_static_property(
```
ptondereau, line 162: so we don't use the `zend_result` here? Is it worth adding error handling here?

src/zend/class.rs
```diff
@@ -132,6 +135,74 @@ impl ClassEntry {
...
+    /// use ext_php_rs::zend::ClassEntry;
+    ///
+    /// let ce = ClassEntry::try_find("MyClass").unwrap();
+    /// ce.set_static_property("counter", 42).unwrap();
+    /// ```
+    pub fn set_static_property<T: IntoZval>(&self, name: &str, value: T) -> Result<()> {
+        let name = CString::new(name)?;
+        let mut zval = value.into_zval(false)?;
+        unsafe {
+            crate::ffi::zend_update_static_property(
```
ptondereau, line 198: same here for `zend_result`

guide/src/macros/classes.md
```diff
@@ -174,3 +178,66 @@ pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
...
+# #![cfg_attr(windows, feature(abi_vectorcall))]
+# extern crate ext_php_rs;
+use ext_php_rs::prelude::*;
+use ext_php_rs::class::RegisteredClass;
+
+#[php_class]
+pub struct Counter {
+    #[php(prop)]
+    pub instance_value: i32,
+    #[php(prop, static)]
```
ptondereau: Do you think it could be difficult to implement a default value here? or does it require a whole implementation for non-static prop?

src/zend/class.rs
ptondereau: (it is only `zend_update_static_property` sorry)

guide/src/macros/classes.md
kakserpom: Good idea. There you go.

PR #632, opened 2025-12-22
feat(zval): Zval coercion

kakserpom:
Rework of https://github.com/extphprs/ext-php-rs/pull/341

### DIFF
--- a/src/types/zval.rs
+++ b/src/types/zval.rs
@@ -4,6 +4,8 @@
 
 use std::{convert::TryInto, ffi::c_void, fmt::Debug, ptr};
 
+use cfg_if::cfg_if;
+
 use crate::types::ZendIterator;
 use crate::types::iterable::Iterable;
 use crate::{
@@ -14,8 +16,8 @@ use crate::{
     error::{Error, Result},
     ffi::{
         _zval_struct__bindgen_ty_1, _zval_struct__bindgen_ty_2, ext_php_rs_zend_string_release,
-        zend_array_dup, zend_is_callable, zend_is_identical, zend_is_iterable, zend_resource,
-        zend_value, zval, zval_ptr_dtor,
+        zend_array_dup, zend_is_callable, zend_is_identical, zend_is_iterable, zend_is_true,
+        zend_resource, zend_value, zval, zval_ptr_dtor,
     },
     flags::DataType,
     flags::ZvalTypeFlags,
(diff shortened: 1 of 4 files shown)


src/types/zval.rs
```diff
@@ -530,6 +532,231 @@ impl Zval {
...
+    /// assert_eq!(zv.coerce_to_bool(), true);
+    /// ```
+    #[must_use]
+    pub fn coerce_to_bool(&self) -> bool {
+        cfg_if! {
+            if #[cfg(php84)] {
+                let ptr: *const Self = self;
+                unsafe { zend_is_true(ptr) }
+            } else {
+                // Pre-PHP 8.5: zend_is_true takes *mut and returns c_int
```
ptondereau:
```suggestion
                // Pre-PHP 8.4: zend_is_true takes *mut and returns c_int
```

It's since PHP 8.4+

src/types/zval.rs
```diff
@@ -947,4 +1268,204 @@ mod tests {
...
+            let null_zv = Zval::null();
+            assert!(!null_zv.coerce_to_bool());
+
+            let empty_array = Zval::new_array();
+            assert!(!empty_array.coerce_to_bool());
+        });
+    }
+
+    #[test]
+    fn test_coerce_to_string() {
```
ptondereau, line 1444: it doesn't support Hexadecimal
PHP's (int)"0xFF" returns 255 and here it returns 0.

Same for Infinity/NaN handling

src/types/zval.rs
```diff
@@ -899,8 +1126,102 @@ impl<'a> FromZvalMut<'a> for &'a mut Zval {
...
+    for c in chars {
+        if c.is_ascii_digit() {
+            num_str.push(c);
+        } else {
+            break;
+        }
+    }
+
+    // Parse or return 0
+    num_str.parse().unwrap_or(0)
```
ptondereau: PHP's `_is_numeric_string_ex` handles overflow by returning the value as `IS_DOUBLE`.
Here, it silently returns 0 on overflow.

src/types/zval.rs
```diff
@@ -899,8 +1126,102 @@ impl<'a> FromZvalMut<'a> for &'a mut Zval {
...
+/// Parses an integer from a string following PHP's type coercion rules.
+///
+/// PHP extracts the leading numeric portion of a string:
+/// - `"42"` → 42
+/// - `"42abc"` → 42
+/// - `"  42"` → 42 (leading whitespace is skipped)
+/// - `"-42"` → -42
+/// - `"abc"` → 0
+/// - `""` → 0
+fn parse_long_from_str(s: &str) -> ZendLong {
```
ptondereau, line 1214: I'd suggest using `is_numeric_string`: it keeps the same behavior as PHP

src/types/zval.rs
kakserpom:
```
➜  ~ php -r 'var_dump((int)"0xFF");'
int(0)
```

ptondereau: Ok I have tested other cases:

- (int)"9999999999999999999999" → 9223372036854775807 (PHP_INT_MAX)
- (int)"-9999999999999999999999" → -9223372036854775808 (PHP_INT_MIN)
- (float)".5" → 0.5
- (float)"-.5" → -0.5
- (string)1e-10 → "1.0E-10" (uppercase E)
- (string)INF → "INF"
- (string)-0.0 → "-0"

kakserpom:
@ptondereau 

> we should keep the same behavior as PHP, so don't hesitate to use internal function

I tried that and it led to some complications across PHP versions. I've decided to stick with Rust. Added more tests.

ptondereau:
> @ptondereau
> 
> > we should keep the same behavior as PHP, so don't hesitate to use internal function
> 
> I tried that and it led to some complications across PHP versions. I've decided to stick with Rust. Added more tests.

ok then it should be documented

ptondereau (approved):

PR #636, opened 2025-12-23
feat(object): Lazy ghost and Lazy Proxy

kakserpom:
https://www.php.net/manual/en/language.oop5.lazy-objects.php

### DIFF
--- a/allowed_bindings.rs
+++ b/allowed_bindings.rs
@@ -107,6 +107,9 @@ bind! {
     zend_hash_str_update,
     zend_internal_arg_info,
     zend_is_callable,
+    zend_is_callable_ex,
+    zend_fcall_info_cache,
+    _zend_fcall_info_cache,
     zend_is_identical,
     zend_is_iterable,
     zend_known_strings,
@@ -266,6 +269,18 @@ bind! {
     zend_std_get_properties,
     zend_std_has_property,
     zend_objects_new,
+    zend_object_make_lazy,
+    zend_lazy_object_init,
+    zend_lazy_object_mark_as_initialized,
+    // Note: zend_lazy_object_get_instance and zend_lazy_object_get_flags are not
+    // exported (no ZEND_API) in PHP, so they cannot be used on Windows.
+    // Use zend_lazy_object_init instead which returns the instance for proxies.
+    zend_class_can_be_lazy,
+    zend_lazy_object_flags_t,
+    ZEND_LAZY_OBJECT_STRATEGY_GHOST,
+    ZEND_LAZY_OBJECT_STRATEGY_PROXY,
+    ZEND_LAZY_OBJECT_INITIALIZED,
+    ZEND_LAZY_OBJECT_SKIP_INITIALIZATION_ON_SERIALIZE,
     zend_standard_class_def,
     zend_class_serialize_deny,
     zend_class_unserialize_deny,
--- a/guide/src/types/object.md
+++ b/guide/src/types/object.md
@@ -78,4 +78,152 @@ pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
 # fn main() {}
 ```
 
+## Lazy Objects (PHP 8.4+)
+
+PHP 8.4 introduced lazy objects, which defer their initialization until their
+properties are first accessed. ext-php-rs provides APIs to introspect and create
+lazy objects from Rust.
+
+### Lazy Object Types
+
+There are two types of lazy objects:
+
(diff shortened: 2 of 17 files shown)


tests/build.rs
```diff
@@ -0,0 +1,88 @@
+//! Build script for the tests crate that detects PHP version and sets cfg flags.
```
ptondereau, line 1: I'm a bit worried about duplicate logic here.
Did you check if the cfg flags are propagated instead? If not, I would suggest first creating another PR about extracting it into a crate like `build` or `php-detect` (or whatever) with extracted logic of `find_php` etc...

allowed_bindings.rs
```diff
@@ -266,6 +269,18 @@ bind! {
     zend_std_get_properties,
     zend_std_has_property,
     zend_objects_new,
+    zend_object_make_lazy,
+    zend_lazy_object_init,
+    zend_lazy_object_mark_as_initialized,
+    // Note: zend_lazy_object_get_instance and zend_lazy_object_get_flags are not
+    // exported (no ZEND_API) in PHP, so they cannot be used on Windows.
+    // Use zend_lazy_object_init instead which returns the instance for proxies.
+    zend_class_can_be_lazy,
```
ptondereau, line 278: Bound but not used

src/types/object.rs
```diff
@@ -351,6 +408,278 @@ impl ZendObject {
...
+        // Get the class entry
+        let ce = self.ce;
+
+        // Make the object lazy
+        let result = unsafe {
+            zend_object_make_lazy(self, ce, &raw mut initializer_zv, &raw mut fcc, strategy)
+        };
+
+        if result.is_null() {
+            Err(Error::InvalidScope)
```
ptondereau: `Error::InvalidScope` is vague IMHO. Maybe adding a dedicated `Error::LazyObjectFailed` variant.

guide/src/types/object.md
```diff
@@ -78,4 +78,136 @@ pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
...
+Available introspection methods:
+
+| Method | Description |
+|--------|-------------|
+| `is_lazy()` | Returns `true` if the object is lazy (ghost or proxy) |
+| `is_lazy_ghost()` | Returns `true` if the object is a lazy ghost |
+| `is_lazy_proxy()` | Returns `true` if the object is a lazy proxy |
+| `is_lazy_initialized()` | Returns `true` if the lazy object has been initialized |
+| `lazy_init()` | Triggers initialization of a lazy object |
+| `lazy_get_instance()` | For proxies, returns the real instance after initialization |
```
ptondereau: But the method signature returns `Option<&mut Self>`, not the instance directly.

tests/build.rs
kakserpom: Done in this PR.

src/types/object.rs
```diff
@@ -351,6 +408,284 @@ impl ZendObject {
...
+    where
+        F: Fn() -> R + 'static,
+        R: IntoZval + 'static,
+    {
+        // Check if the class can be made lazy
+        let ce = unsafe { self.ce.as_ref() }.ok_or(Error::InvalidPointer)?;
+        if !ce.can_be_lazy() {
+            return Err(Error::LazyObjectFailed);
+        }
+
```
ptondereau, line 647: we should also add 
```rust
if self.is_lazy() && !self.is_lazy_initialized() {
  return Err(Error::LazyObjectFailed);
}
```
this cover: https://github.com/php/php-src/blob/fb5d4784fe56ffd1ac11b227f7859e1f37d0207c/Zend/zend_lazy_objects.c#L233

windows_build.rs
```diff
@@ -34,7 +34,10 @@ impl<'a> PHPProvider<'a> for Provider<'a> {
     fn new(info: &'a PHPInfo) -> Result<Self> {
         let version = info.version()?;
         let is_zts = info.thread_safety()?;
-        let arch = info.architecture()?;
+        let arch: Arch = info
+            .get_key("Architecture")
+            .context("Could not find architecture of PHP")?
+            .try_into()?;
```
ptondereau: why this change?

kakserpom: https://github.com/extphprs/ext-php-rs/actions/runs/21292094084/job/61288125843?pr=636

ptondereau: I think this is because it hasn't been backported

kakserpom:
@ptondereau All green now

PR #639, opened 2025-12-25
feat(class): `readonly` and `final` classes

kakserpom:


crates/macros/src/class.rs
```diff
@@ -198,9 +204,40 @@ fn generate_registered_class_impl(
         }
     });
 
-    let flags = match flags {
-        Some(flags) => flags.to_token_stream(),
-        None => quote! { ::ext_php_rs::flags::ClassFlags::empty() }.to_token_stream(),
+    // Generate flags expression, combining user-provided flags with readonly if
+    // specified. Note: ReadonlyClass is only available on PHP 8.2+, so we use
+    // cfg to conditionally include it.
```
Xenira: We should probably fail to compile or emmit a warning if readonly is set as it would otherwise be dropped silently

kakserpom: Done.

tests/build.rs
```diff
@@ -0,0 +1,88 @@
...
+        if minor >= 3 {
+            println!("cargo:rustc-cfg=php83");
+        }
+        if minor >= 4 {
+            println!("cargo:rustc-cfg=php84");
+        }
+        if minor >= 5 {
+            println!("cargo:rustc-cfg=php85");
+        }
+    }
```
Xenira: `ext_php_rs::build::register_php_cfg_flags();`

Cargo.toml
```diff
@@ -4,6 +4,7 @@ description = "Bindings for the Zend API to build PHP extensions natively in Rus
 repository = "https://github.com/extphprs/ext-php-rs"
 homepage = "https://ext-php.rs"
 license = "MIT OR Apache-2.0"
+links = "ext-php-rs"
```
Xenira: linking with itself seems wrong. Havent looked into it though.

kakserpom: Without `links`, the DEP_* variables wouldn't be available to dependent crates.

tests/build.rs
kakserpom: @Xenira Could you take a look?

kakserpom:
@Xenira @ptondereau All is rebased and green.

src/lib.rs
```diff
@@ -72,6 +72,21 @@ pub const PHP_DEBUG: bool = cfg!(php_debug);
...
+pub const PHP_82: bool = cfg!(php82);
+
+/// Whether the extension is compiled for PHP 8.3 or later.
+pub const PHP_83: bool = cfg!(php83);
+
+/// Whether the extension is compiled for PHP 8.4 or later.
+pub const PHP_84: bool = cfg!(php84);
+
+/// Whether the extension is compiled for PHP 8.5 or later.
+pub const PHP_85: bool = cfg!(php85);
```
ptondereau, line 88: Are they being used anywhere? If not, let's remove it

crates/macros/src/class.rs
```diff
@@ -198,9 +204,46 @@ fn generate_registered_class_impl(
...
+            // Only readonly flag
+            quote! {
+                {
+                    #[cfg(not(php82))]
+                    compile_error!("The `readonly` class attribute requires PHP 8.2 or later");
+
+                    #[cfg(php82)]
+                    { ::ext_php_rs::flags::ClassFlags::ReadonlyClass }
+                    #[cfg(not(php82))]
+                    { ::ext_php_rs::flags::ClassFlags::empty() }
```
ptondereau, line 242: harmless, but this branch will never be reached

kakserpom: It's for type-checking

src/lib.rs
kakserpom: These are public and can be used in the extension code.

PR #640, opened 2026-01-12
fix(deps): bump parking_lot required version to 0.12.3

TobiasBengtsson:
to include support for parking_lot::ArcRwLockReadGuard and parking_lot::ArcRwLockWriteGuard that we use in src/zend/globals.rs, that was only publicly exposed in 0.12.3 (commit [[c357017d](https://github.com/Amanieu/parking_lot/commit/c357017decfa0f21ca597a4958745ad0999cf9eb)]).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ autotests = false
 
 [dependencies]
 bitflags = "2"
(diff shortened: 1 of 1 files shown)


Xenira:
LGTM thanks

PR #643, opened 2026-01-14
feat(string): smartstring support

kakserpom:


src/types/string/smartstring_impl.rs
```diff
@@ -0,0 +1,50 @@
...
+impl<M: smartstring::SmartStringMode> IntoZval for SmartString<M> {
+    const TYPE: DataType = DataType::String;
+    const NULLABLE: bool = false;
+
+    fn set_zval(self, zv: &mut Zval, persistent: bool) -> Result<()> {
+        zv.set_string(self.as_str(), persistent)
+    }
+}
+
+impl FromZval<'_> for SmartString<LazyCompact> {
```
ptondereau: Should we be more generic here?
```suggestion
impl<M: smartstring::SmartStringMode> FromZval<'_> for SmartString<M> {
```

kakserpom: Thanks

PR #654, opened 2026-01-27
docs(guide): customize favicon

c14n:
Every piece of documentation built with the `mdbook` tool that uses the default theme also uses the default favicon, which makes it hard to recognize the correct browser tab for the particular piece of documentation you're looking for if you're a tab hoarder like myself.

So I just pilfered the logo from the org's profile, rescaled it, and placed it in the location that is picked up by `mdbook` .

It's not a perfect solution. The logo has too much detail to be easily recognizable as a crabephant (elephtacean? pinceratrunkus?) at smaller sizes, but it's different enough from the default to be noticeable, which solves the problem.

## Checklist

- [x] I have added tests that prove my code works as expected. **(N/A)**
- [x] I have added documentation if applicable. **(N/A)**
- [x] I have added a [migration guide](../CONTRIBUTING.md#breaking-changes) if applicable. **(N/A)**


ptondereau:
hello and thanks! Still a good contrib

PR #656, opened 2026-01-27
feat(class): abstract and final methods

kakserpom:
## Description

Brings about abstract and final methods defined in Rust.

### DIFF
--- a/crates/macros/src/impl_.rs
+++ b/crates/macros/src/impl_.rs
@@ -75,6 +75,8 @@ struct MethodArgs {
     vis: Visibility,
     /// Method type.
     ty: MethodTy,
+    /// Whether this is a final method.
+    is_final: bool,
 }
 
 #[derive(FromAttributes, Default, Debug)]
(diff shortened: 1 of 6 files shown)


crates/macros/src/impl_.rs
```diff
@@ -242,11 +252,21 @@ impl<'a> ParsedImpl<'a> {
...
+                            modifiers.insert(MethodModifier::Final);
                         }
 
-                        let builder = func.function_builder(&call_type);
+                        // Abstract methods use a different builder that doesn't generate a handler
+                        let builder = if is_abstract {
+                            func.abstract_function_builder()
+                        } else {
+                            func.function_builder(&call_type)
+                        };
```
ptondereau, line 311: be aware of these cases:

```rust
if is_abstract && opts.is_final {
      bail!(method.sig.ident => "Methods cannot be both abstract and final.");
}
if is_abstract && matches!(opts.vis, Visibility::Private) {
      bail!(method.sig.ident => "Abstract methods cannot be private.");
}
```

PR #662, opened 2026-01-29
fix(stubs): Proper stub generation for interfaces

kakserpom:


ptondereau:
as we don't have tests for cargo-php do you have any before/after output?

kakserpom:
1) `cargo php stubs` crashed because of dtor calls.

2) Interface stubs were not generated

My use case:
https://github.com/kakserpom/php-sqlx-rs/tree/master

crates/macros/src/impl_interface.rs
```diff
@@ -146,6 +146,16 @@ fn generate_method_builder(
     output: &ReturnType,
     is_static: bool,
 ) -> TokenStream {
+    // Helper to check if a type is Option<T>
+    fn is_option_type(ty: &syn::Type) -> bool {
```
ptondereau, line 150: This assumes that all optional arguments are at the end of the function's signature.
What if we have: `fn foo(&self, opt: Option<i64>, required: i64)`?

src/types/zval.rs
```diff
@@ -756,15 +756,22 @@ impl Zval {
...
+        // This is important for cargo-php stub generation where zval_ptr_dtor is
+        // stubbed as a null pointer - calling it would crash.
         if self.is_string() {
             unsafe {
                 if let Some(str_ptr) = self.value.str_.as_mut() {
                     ext_php_rs_zend_string_release(str_ptr);
                 }
             }
-        } else {
+        } else if self.is_array() || self.is_object() || self.is_resource() || self.is_reference() {
```
ptondereau:
```suggestion
        } else if self.is_array() || self.is_object() || self.is_resource() || self.is_reference() || self.get_type() == DataType::ConstantExpression {
```

`IS_CONSTANT_AST` needs to be destroyed too.

But maybe a cleaner way is to add a method `is_constant_ast()`

PR #665, opened 2026-02-02
feat(class): allow implementing internal PHP interfaces

kakserpom:


ptondereau:
What is the point of exposing it? As we have implement support

kakserpom:
> What is the point of exposing it? As we have implement support
```
#[php(implements(ce = ce::jsonserializable, stub = "\\JsonSerializable"))]
```

ptondereau:
> > What is the point of exposing it? As we have implement support
> ```
> #[php(implements(ce = ce::jsonserializable, stub = "\\JsonSerializable"))]
> ```

Ok that means that for every internal PHP class like this one or DateTimeInterface for example we need to re-expose it?

kakserpom:
> Ok that means that for every internal PHP class like this one or DateTimeInterface for example we need to re-expose it?

I guess, if it's exported in Zend API.

ptondereau:
so is there any way to be able to do a `#[php(implement("\\JsonSerializable")]` ?

kakserpom:
> so is there any way to be able to do a `#[php(implement("\\JsonSerializable")]` ?

Done.

ptondereau:
- let's add some tests.
- update PR title

kakserpom:
@ptondereau I cannot edit the PR title. I am not worthy :-D

kakserpom:
@ptondereau Thanks. I've just changed the approach a little bit and added some tests.

crates/macros/src/class.rs
```diff
@@ -105,6 +118,20 @@ impl ToTokens for ClassEntryAttribute {
...
+                // For a string literal, generate a function that looks up the class at runtime
+                // Uses try_find_no_autoload to avoid triggering autoloading during MINIT
+                // Strip leading backslash since PHP's class table stores names without it
+                // Convert to lowercase since PHP's class table uses lowercase keys
+                let lookup_name = name.strip_prefix('\\').unwrap_or(name).to_lowercase();
+                quote! {
+                    (
+                        || ::ext_php_rs::zend::ClassEntry::try_find_no_autoload(#lookup_name)
+                            .expect(concat!("Failed to find class entry for ", #lookup_name)),
+                        #name
```
ptondereau, line 131:
```suggestion
                        || ::ext_php_rs::zend::ClassEntry::try_find_no_autoload(#lookup_name)
                            .expect(concat!("Failed to find class entry for ", #name)),
```

src/zend/class.rs
```diff
@@ -35,6 +35,39 @@ impl ClassEntry {
...
+        use crate::types::ZendHashTable;
+        use crate::zend::CompilerGlobals;
+
+        let cg = CompilerGlobals::get();
+        let class_table = cg.class_table()?;
+
+        // zend_hash_str_find_ptr_lc handles lowercase conversion internally
+        let ce_ptr = unsafe {
+            crate::ffi::zend_hash_str_find_ptr_lc(
+                ptr::from_ref::<ZendHashTable>(class_table).cast(),
```
ptondereau: could you try to remove `.cast()` here?

crates/macros/src/class.rs
```diff
@@ -105,6 +118,20 @@ impl ToTokens for ClassEntryAttribute {
                     )
                 }
             }
+            ClassEntryAttribute::Name(name) => {
+                // For a string literal, generate a function that looks up the class at runtime
+                // Uses try_find_no_autoload to avoid triggering autoloading during MINIT
+                // Strip leading backslash since PHP's class table stores names without it
+                // Convert to lowercase since PHP's class table uses lowercase keys
+                let lookup_name = name.strip_prefix('\\').unwrap_or(name).to_lowercase();
```
ptondereau: Looking at `zend_hash_str_find_ptr_lc` it already transforms it to lowercase

src/zend/mod.rs
```diff
@@ -38,6 +38,7 @@ pub use ex::ExecuteData;
 pub use exception_observer::{ExceptionInfo, ExceptionObserver};
 pub use function::Function;
 pub use function::FunctionEntry;
+pub use globals::CompilerGlobals;
```
ptondereau: Is it really necessary? what about pub(crate)?

crates/macros/src/class.rs
```diff
@@ -44,13 +45,20 @@ pub struct StructAttributes {
...
+/// #[php(implements("\\JsonSerializable"))]
+/// ```
 #[derive(Debug)]
 pub enum ClassEntryAttribute {
     /// Explicit class entry with a function returning `&'static ClassEntry` and
     /// a stub name
     Explicit { ce: syn::Expr, stub: String },
     /// A Rust type that implements `RegisteredClass`
     Type(syn::Path),
+    /// A PHP class name to be looked up at runtime via `ClassEntry::try_find`
```
ptondereau:
```suggestion
    /// A PHP class name to be looked up at runtime via `ClassEntry::try_find_no_autoload`
```

Issue #182, opened 2022-11-13
class structs cannot contain class struct properties


ju1ius:
Hi, currently the following is not possible:
```rust
#[php_class]
struct Foo {
  #[prop]
  bar: Bar,
}

#[php_class]
#[derive(Clone)]
struct Bar {
  #[prop]
  value: String,
}
```
```
the trait bound `Bar: FromZval<'_>` is not satisfied
the trait `FromZval<'a>` is implemented for `&'a Bar`
required for `Bar` to implement `Prop<'_>`
required for the cast from `Bar` to the object type `dyn Prop<'_>` rustc(E0277)
```

This is supported in PyO3, from which this library seems to be inspired, so we should be able to do the same in ext_php_rs:
```rust
#[pyclass]
struct Foo {
  #[pyo3(get)]
  bar: Bar,
}

#[pyclass]
#[derive(Clone)]
struct Bar {
  #[pyo3(get)]
  value: String,
}
```

kakserpom:
Any updates? @davidcole1340

Xenira:
@kakserpom david is no longer active here. I am currently one of the main maintainers.

I haven't had time to look into this, as there is currently a big breaking change in the works. But I will have a look at this, just can't promise a timeframe.

PR #668, opened 2026-02-04
feat(class): Class structs can now contain other class structs #182

kakserpom:
#182

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -395,6 +395,67 @@ macro_rules! class_derives {
     };
 }
 
+/// Derives additional traits for cloneable [`RegisteredClass`] types to enable
+/// using them as properties of other `#[php_class]` structs.
+///
+/// This macro should be called for any `#[php_class]` struct that:
+/// 1. Implements [`Clone`]
+/// 2. Needs to be used as a property in another `#[php_class]` struct
+///
+/// The macro implements [`FromZendObject`] and [`FromZval`] for the owned type,
+/// allowing PHP objects to be cloned into Rust values.
+///
+/// # Example
+///
+/// ```ignore
+/// use ext_php_rs::prelude::*;
+/// use ext_php_rs::class_derives_clone;
+///
+/// #[php_class]
+/// #[derive(Clone)]
+/// struct Bar {
+///     #[php(prop)]
+///     value: String,
+/// }
+///
+/// class_derives_clone!(Bar);
+///
+/// #[php_class]
+/// struct Foo {
+///     #[php(prop)]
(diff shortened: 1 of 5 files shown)


kakserpom:
CI is failing spuriously again :-(

src/macros.rs
```diff
@@ -333,6 +333,67 @@ macro_rules! class_derives {
...
+/// [`RegisteredClass`]: crate::class::RegisteredClass
+/// [`FromZendObject`]: crate::convert::FromZendObject
+/// [`FromZval`]: crate::convert::FromZval
+#[macro_export]
+macro_rules! class_derives_clone {
+    ($type: ty) => {
+        impl $crate::convert::FromZendObject<'_> for $type {
+            fn from_zend_object(obj: &$crate::types::ZendObject) -> $crate::error::Result<Self> {
+                let class_obj = $crate::types::ZendClassObject::<$type>::from_zend_obj(obj)
+                    .ok_or($crate::error::Error::InvalidScope)?;
```
ptondereau, line 441: `InvalidScope` is used when the `ZendClassObject` can't be extracted. We should create a dedicated error for that case.

src/macros.rs
```diff
@@ -333,6 +333,67 @@ macro_rules! class_derives {
...
+                Ok((**class_obj).clone())
```
ptondereau, line 442: So we are cloning here, but in the PHP tests we're only verifying that reading works; that means only reading is safe and not writing (we're updating the clone object and not the original).

It's ok to me, but should be documented then

ptondereau: + What if a subclass contains `Rc`/`Arc`

tests/src/integration/class/class.php
```diff
@@ -348,3 +348,20 @@ public function __construct(string $data) {
...
+assert($outer->getInnerValue() === 'hello world', 'OuterClass should be able to access inner value');
+
+// Test that the inner property is properly accessible
+assert($outer->inner instanceof InnerClass, 'outer->inner should be InnerClass instance');
+assert($outer->inner->value === 'hello world', 'outer->inner->value should be accessible');
+
+// Test setting inner property
+$newInner = new InnerClass('new value');
+$outer->inner = $newInner;
+assert($outer->getInnerValue() === 'new value', 'After setting inner, value should be updated');
```
ptondereau, line 367: I would add a test for asserting clone-on-read behavior:

```php
$a = $outer->inner;
$b = $outer->inner;
assert($a !== $b);
$a->value = 'changed';
assert($outer->inner->value === 'original');
```

Issue #522, opened 2025-07-17
Add `indexmap` feature


Xenira:
## Description
To preserve insert order we should add `indexmap` as an optional feature. 

## Use Case
Only `Vec<(K, V)>` can currently be used to have this feature and you loose key access with that.

### Additional Context

- [ ] I can try implementing this feature myself (please assign me to this issue)

PR #670, opened 2026-02-05
feat(types): indexmap feature #522

kakserpom:
#522

### DIFF
--- a/guide/src/types/hashmap.md
+++ b/guide/src/types/hashmap.md
@@ -3,7 +3,7 @@
 `HashMap`s are represented as associative arrays in PHP.
 
 | `T` parameter | `&T` parameter | `T` Return type | `&T` Return type | PHP representation |
-| ------------- | -------------- | --------------- | ---------------- | ------------------ |
+|---------------|----------------|-----------------|------------------|--------------------|
 | Yes           | No             | Yes             | No               | `ZendHashTable`    |
 
 Converting from a zval to a `HashMap` is valid when the key is a `String`, and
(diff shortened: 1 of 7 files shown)


ptondereau:
Don't forget to add the feature flags in the build.yml because it won't load your tests

src/types/array/conversions/index_map.rs
```diff
@@ -0,0 +1,337 @@
...
+    const TYPE: DataType = DataType::Array;
+    const NULLABLE: bool = false;
+
+    fn set_zval(self, zv: &mut Zval, _: bool) -> Result<()> {
+        let arr = self.try_into()?;
+        zv.set_hashtable(arr);
+        Ok(())
+    }
+}
+
```
ptondereau, line 123: can we also add support for `IndexSet` like you've done with `HashSet`

guide/src/types/hashmap.md
```diff
@@ -16,13 +16,13 @@ Converting from a `HashMap` to a zval is valid when the key implements
 
 <div class="warning">
 
-    When using `HashMap` the order of the elements it not preserved.
+    When using `HashMap` the order of the elements is not preserved.
```
ptondereau, line 19: Technically, BTreeMap also supports order preservation (sorted order vs insert order)

kakserpom: BTreeMap doesn't preserve order, it enforces order by its nature.

ptondereau (approved):

PR #671, opened 2026-02-05
feat: eval PHP code from files

ptondereau:
## Description

PHP extensions often need to run PHP code at startup for bootstrapping tasks like registering autoloaders, defining helper functions, and configuring error handlers. We have no way to do this. The existing embed module is for the embed SAPI (running PHP from a standalone Rust binary), not for executing PHP snippets from within a loaded extension.

This adds `php_eval::execute()` so extension authors can embed .php files with `include_bytes!` and run them from any lifecycle hook they choose.

### DIFF
--- a/src/php_eval.rs
+++ b/src/php_eval.rs
@@ -1,0 +1,322 @@
+//! Execute embedded PHP code within a running PHP extension.
+//!
+//! This module provides a way to compile and execute PHP code that has been
+//! embedded into the extension binary at compile time using `include_bytes!`.
+//!
+//! Uses `zend_compile_string` + `zend_execute` (not `zend_eval_string`)
+//! to avoid security scanner false positives and compatibility issues
+//! with hardened PHP configurations.
+//!
+//! # Example
+//!
+//! ```rust,ignore
+//! use ext_php_rs::php_eval;
+//!
+//! // Both include_bytes! and include_str! are supported:
+//! const SETUP_BYTES: &[u8] = include_bytes!("../php/setup.php");
+//! const SETUP_STR: &str = include_str!("../php/setup.php");
+//!
+//! php_eval::execute(SETUP_BYTES).expect("failed to execute embedded PHP");
+//! php_eval::execute(SETUP_STR).expect("failed to execute embedded PHP");
+//! ```
+
+use crate::ffi;
+use crate::types::ZendStr;
+use crate::zend::try_catch;
+use std::fmt;
+use std::mem;
+use std::panic::AssertUnwindSafe;
+
+/// Errors that can occur when executing embedded PHP code.
+#[derive(Debug)]
+pub enum PhpEvalError {
+    /// The code does not start with a `<?php` open tag.
+    MissingOpenTag,
+    /// PHP failed to compile the code (syntax error).
+    CompilationFailed,
+    /// The code executed but threw an unhandled exception.
+    ExecutionFailed,
+    /// A PHP fatal error (bailout) occurred during execution.
+    Bailout,
+}
+
+impl fmt::Display for PhpEvalError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            PhpEvalError::MissingOpenTag => {
+                write!(f, "PHP code must start with a <?php open tag")
+            }
+            PhpEvalError::CompilationFailed => write!(f, "PHP compilation failed (syntax error)"),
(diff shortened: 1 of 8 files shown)


src/php_eval.rs
```diff
@@ -0,0 +1,218 @@
...
+fn strip_php_open_tag(code: &[u8]) -> &[u8] {
+    let trimmed = match code
+        .iter()
+        .position(|&b| b != b' ' && b != b'\t' && b != b'\r' && b != b'\n')
+    {
+        Some(pos) => &code[pos..],
+        None => return code,
+    };
+
+    if trimmed.starts_with(b"<?php") {
```
kakserpom: Should be case insensitive

src/php_eval.rs
```diff
@@ -0,0 +1,218 @@
...
+///
+/// * `code` - Raw PHP source bytes, typically from `include_bytes!`.
+///
+/// # Errors
+///
+/// Returns [`PhpEvalError`] if compilation fails, an exception is thrown,
+/// or a fatal error occurs.
+pub fn execute(code: &[u8]) -> Result<(), PhpEvalError> {
+    let code = strip_bom(code);
+    let code = strip_php_open_tag(code);
```
kakserpom: Alternatively we could prepend `?>`

src/php_eval.rs
```diff
@@ -0,0 +1,218 @@
...
+        &code[3..]
+    } else {
+        code
+    }
+}
+
+fn strip_php_open_tag(code: &[u8]) -> &[u8] {
+    let trimmed = match code
+        .iter()
+        .position(|&b| b != b' ' && b != b'\t' && b != b'\r' && b != b'\n')
```
kakserpom: is_ascii_whitespace

src/php_eval.rs
```diff
@@ -0,0 +1,218 @@
...
+    {
+        Some(pos) => &code[pos..],
+        None => return code,
+    };
+
+    if trimmed.starts_with(b"<?php") {
+        let rest = &trimmed[5..];
+        if rest.starts_with(b"\r\n") {
```
kakserpom: Why not just trim_start?

src/php_eval.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    if code.is_empty() {
+        return Ok(());
+    }
+
+    let source = ZendStr::new(code, false);
+
+    let result = try_catch(AssertUnwindSafe(|| unsafe {
+        let eg = ffi::ext_php_rs_executor_globals();
+        let prev_error_reporting = (*eg).error_reporting;
```
kakserpom: mem::swap?

src/php_eval.rs
ptondereau: Making this case-insensitive would be more permissive than PHP itself. I think users will match this thanks to the ide/lsp/etc...

src/php_eval.rs
ptondereau: `error_reporting` is a plain integer (Copy type), so `mem::swap` adds nothing.

src/php_eval.rs
ptondereau: It adds an unnecessary mode switch for me, and it would cause the scanner to treat the rest as HTML, not PHP.

src/php_eval.rs
kakserpom:
```
➜  ~ php -r '?>Hello<?PHP echo " World!".PHP_EOL;'
Hello World!
```

src/php_eval.rs
kakserpom: Well I don't really understand the idea here. This makes `<?php foo();` and `foo();` behave identically.  

I understand that `ext_php_rs_zend_compile_string()` expects PHP code without `<?php`, but if you prepend `?>` if will handle `<?php` just fine.

We should either prohibit `<?php` at the start or make it mandatory.

src/php_eval.rs
kakserpom: `let prev_error_reporting = mem::take(&mut (*eg).error_reporting);`

? :-)

src/php_eval.rs
ptondereau: Ok I've made it mandatory then

kakserpom:
Why include_bytes instead of include_str?

ptondereau:
> Why include_bytes instead of include_str?

This is because PHP files are not guaranteed to be UTF-8 and I pass raw bytes directly to `zend_compile_string` and take care of the validation

Issue #376, opened 2025-02-25
[stubs] Property is missing `ty`, `default` and `nullable`


Xenira:
https://github.com/davidcole1340/ext-php-rs/blob/7d23765294984340084757fd83a8232474541eff/crates/macros/src/module.rs#L273

Xenira:
As this is not a regression I will remove this from the milestone.

kakserpom:
https://github.com/extphprs/ext-php-rs/pull/673

PR #673, opened 2026-02-08
fix(stubs): Missing property types #184

### LINKED ISSUE #184: [stubs] missing type declarations for properties
| Component | Version |
| :---              |     :---:    | 
| ext_php_rs  | git:master |
| cargo-php  |  v0.1.7 |

Extension code:
```rust
#[php_class]
pub struct Example {
  #[prop] pub foo: String,
  #[prop] pub bar: Option<String>,
}

#[php_impl]
impl Example {
  #[getter]
  pub fn get_baz(&self) -> Vec<bool> {
    vec![true, false]
  }
}

#[php_startup]
pub fn startup_function() {}

#[php_module]
    module
}
```

Expected stubs:
```php
<?php

class Example {
  public string $foo;
  public string|null $bar;
  /**
   * @var array<bool>
   */
  public readonly array $baz;
}
```

Actual stubs:
```php

<?php
class Example {
  public $foo;
  public $bar;
  public $baz;
  public function getBaz(): array {}
}
```

kakserpom:
Fixes #184 and #376

### DIFF
--- a/crates/cli/tests/snapshots/stubs_snapshot__hello_world_stubs.snap
+++ b/crates/cli/tests/snapshots/stubs_snapshot__hello_world_stubs.snap
@@ -12,52 +12,74 @@ namespace {
 
     const HELLO_WORLD = 100;
 
+    /**
+     * A simple test class demonstrating ext-php-rs features.
+     *
+     * This class showcases property definitions, constants, and various
+     * method types including constructors and static methods.
+     */
     class TestClass {
         const NEW_CONSTANT_NAME = 5;
 
         const SOME_OTHER_STR = 'Hello, world!';
 
-        public $a;
+        public int $a;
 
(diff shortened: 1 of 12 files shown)


crates/cli/tests/snapshots/stubs_snapshot__hello_world_stubs.snap
```diff
@@ -17,9 +17,9 @@ namespace {
```
ptondereau, line 23: Don't hesitate to update the example. It also acts as a showcase for the whole crate.

ptondereau:
Hey @kakserpom, thanks for working on this! We ended up merging #705 which covers the same ground (fixes #184 and #376). Closing this one — appreciate the effort and sorry for the overlap!

PR #675, opened 2026-02-10
chore(tests): Test to close #176

### LINKED ISSUE #176: Don't expose accessor methods when using getter/setter macros
When using the getter/setter macros, the property is created, but the accessor methods are exposed to php.
Also, the generated property is not exposed to the Reflection API.

```rust
#[php_class]
pub struct Test(u64);

#[php_impl]
impl Test {
  pub fn __construct(field: u64) { Self(field) }
  #[getter]
  pub fn get_field() -> u64 { self.0 }
}
```

```php
<?php
$test = new Test(42);
var_dump($test->field);
$rc = new ReflectionClass($test);
var_dump($rc);
var_dump($rc->getMethods());
```
Expected result:
```
int(42)
array(1) {
  [0] =>
  class ReflectionProperty#4 (2) {
    public string $name =>
    string(8) "field"
    public string $class =>
    string(7) "Test"
  }
}
array(1) {
  [0] =>
  class ReflectionMethod#5 (2) {
    public string $name =>
    string(11) "__construct"
    public string $class =>
    string(3) "Test"
  }
}
```
Actual result:
```
int(42)
array(0) {
}
array(2) {
  [0] =>
  class ReflectionMethod#3 (2) {
    public string $name =>
    string(6) "getField"
    public string $class =>
    string(3) "Foo"
  }
  [1] =>
  class ReflectionMethod#5 (2) {
    public string $name =>
    string(11) "__construct"
    public string $class =>
    string(3) "Test"
  }
}
```

kakserpom:
Closes #176

### DIFF
--- a/tests/src/integration/class/class.php
+++ b/tests/src/integration/class/class.php
@@ -33,6 +33,17 @@ assert($class->booleanProp);
 $class->booleanProp = false;
 assert($class->booleanProp === false);
 
+// Test Issue #176 - getter/setter methods should NOT be exposed, properties SHOULD be
+$testClassReflection = new ReflectionClass(TestClass::class);
+$methodNames = array_map(fn($m) => $m->getName(), $testClassReflection->getMethods());
+$propertyNames = array_map(fn($p) => $p->getName(), $testClassReflection->getProperties());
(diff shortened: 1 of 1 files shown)


ptondereau:
Please remove the useless comment or add to your CLAUDE.md "Don’t comment code unless you are explaining why an unusual pattern was followed."

PR #678, opened 2026-02-13
feat(observer): allow zend_extension registration for low level function

ptondereau:
## Description

Closes the API gap for building `zend_extension`-style profilers (APM, coverage tools, line-level timers) on top of ext-php-rs. Until now the crate only exposed the Observer API (function-call level). This PR adds dual registration so a single extension can also hook the engine's compile-time and per-statement callbacks.

### What you get

```rust
module
    .zend_extension(|| MyProfiler::new())
    .hook_statements()
    .hook_fcalls()
    .finish()
```

The handler trait carries all hooks with default no-op implementations; the builder decides which ones the engine will actually call. Enabling a hook also flips the matching PHP compiler flag so the right opcodes get emitted.

### Why the opt-in

`hook_statements()` and `hook_fcalls()` tell the PHP engine to emit extra opcodes in every compiled script (`ZEND_EXT_STMT`, `ZEND_EXT_FCALL_BEGIN`/`END`). Paying that tax by default would slow every PHP script, even when your profiler doesn't need the data. The builder makes the trade-off explicit. `hook_op_array_compile()` has no compile-time cost because its flag is already set by PHP's defaults; it only controls whether the dispatcher is registered.

### ZTS

Flags are re-asserted in `on_activate`, which is per-request-per-thread, so worker threads created after MINIT still get them. Scripts pre-compiled by opcache before a thread's first activation may miss hooks; load the extension via `zend_extension=...` in `php.ini` if you need 100% coverage there.

### Docs

Guide section at `guide/src/advanced/observer.md#zend-extension-handler` rewritten. Example at `examples/zend_extension.rs` updated. All trait methods renamed to `on_*` prefix for consistency with sibling observers.

## Checklist

- [x] I have read the contribution guidelines and the code of conduct.
- [x] I have added a migration guide if applicable. (N/A: the earlier `zend_extension_handler(factory)` API shipped only on this unmerged branch, no release notes or published docs referenced it.)

### DIFF
--- a/src/zend/zend_extension.rs
+++ b/src/zend/zend_extension.rs
@@ -1,0 +1,499 @@
+//! Zend Extension API bindings for low-level engine hooks.
+//!
+//! Enables building profilers, APMs, and code coverage tools by registering
+//! as a `zend_extension` alongside the regular PHP extension.
+//!
+//! # Example
+//!
+//! ```ignore
+//! use ext_php_rs::prelude::*;
+//! use ext_php_rs::ffi::zend_op_array;
+//!
+//! struct MyProfiler;
+//!
+//! impl ZendExtensionHandler for MyProfiler {
+//!     fn on_op_array_compiled(&self, _op_array: &mut zend_op_array) {}
+//!     fn on_statement(&self, _execute_data: &ExecuteData) {}
+//! }
+//! ```
+
+use std::ffi::{CString, c_void};
+use std::sync::OnceLock;
+
+use crate::ffi;
+use crate::zend::ExecuteData;
+
+/// Trait for handling low-level Zend Engine hooks via `zend_extension`.
+///
+/// All methods have default no-op implementations. Override only what you need.
+///
+/// # Thread Safety
+///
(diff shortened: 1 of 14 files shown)


src/zend/zend_extension.rs
```diff
@@ -0,0 +1,269 @@
...
+    ext.shutdown = None;
+    ext.activate = Some(ext_activate);
+    ext.deactivate = Some(ext_deactivate);
+    ext.message_handler = Some(ext_message_handler);
+    ext.op_array_handler = Some(ext_op_array_handler);
+    ext.statement_handler = Some(ext_statement_handler);
+    ext.fcall_begin_handler = Some(ext_fcall_begin_handler);
+    ext.fcall_end_handler = Some(ext_fcall_end_handler);
+    ext.op_array_ctor = Some(ext_op_array_ctor);
+    ext.op_array_dtor = Some(ext_op_array_dtor);
```
ptondereau, line 356: maybe adding a builder

PR #681, opened 2026-02-17
fix(class): Constructors with subclasses #138

### LINKED ISSUE #138: Constructors with subclasses

kakserpom:
Fixes #138


kakserpom:
@ptondereau The CI is acting up again :-)

Issue #102, opened 2021-10-07
Add support for references


davidcole1340:
Should be able to accept a reference inside a zval: https://www.phpinternalsbook.com/php7/zvals/references.html

Questions to be answered:
- What do we do if we attempt to extract a string from a zval containing a reference? Dereference the reference and attempt to get the string?

ptondereau:
This was addressed in #706 with the introduction of `PhpRef<'a>` — a dedicated wrapper type for PHP references (`&$x`).

`PhpRef` is the explicit opt-in for `ZEND_SEND_BY_REF`. The existing `dereference_mut()` in the arg extraction path already unwraps the `zend_reference` before passing to `FromZvalMut`, so extracting typed values (strings, arrays, etc.) from reference-wrapped zvals works transparently.

See the [guide page](https://extphprs.github.io/ext-php-rs/types/zval_ownership.html) for usage.

PR #682, opened 2026-02-18
fix(zval): Pass-by-reference tests and minor refactoring #102

kakserpom:
#102

### DIFF
--- a/src/describe/abi.rs
+++ b/src/describe/abi.rs
@@ -116,6 +116,7 @@ impl RString {
     /// # Panics
     ///
     /// * If the string is not valid UTF-8
+    #[inline]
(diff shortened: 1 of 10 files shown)


ptondereau:
Thanks for the work here @kakserpom! The safe parts of this PR (dereference_mut optimization, #[inline] annotations, reference tests) have been cherry-picked and merged in #708, with accessor auto-deref added using PHP's fast-path pattern to avoid the overhead on non-references.

The key change vs this PR: we auto-deref in accessors only (not in `is_*()` predicates) to avoid a +35% instruction regression on type checks. See #708 for the full analysis.

Closing in favor of #708.

PR #690, opened 2026-03-17
doc(showcase): add php-hardened and php-sqlx

kakserpom:


ptondereau:
feel free to submit here now: https://github.com/extphprs/awesome-ext-php-rs

PR #720, opened 2026-04-14
feat: static property descriptors, zero heap allocation

### LINKED ISSUE #70: Store properties in `ClassMetadata`

ptondereau:
## Description

Closes #70

The property system used to heap-allocate closures (`Box<dyn Fn>`) and a `HashMap` for every PHP class at runtime. All of this metadata is known at compile time, so there's no reason for it to live on the heap.

This replaces the whole thing with static function pointers and compile-time arrays. Property lookup becomes a linear scan over two small slices (field props first, then getter/setter props), which beats hashing for the typical class size.

**What changed:**
- New `PropertyDescriptor<T>` type with `fn(&T, &mut Zval)` / `fn(&mut T, &Zval)` pointers
- `ClassMetadata` holds `&'static [PropertyDescriptor<T>]` instead of `OnceCell<HashMap<...>>`
- Macros generate concrete getter/setter functions and static arrays
- Removed `Property` enum, `Prop` trait, `PropertyInfo` struct, and all the boxing machinery
- Added callgrind-based property read/write benchmarks

**Benchmark results (callgrind, 100k iterations x 4 properties):**

| Operation | Before (instructions) | After | Improvement |
|---|---|---|---|
| Read | 235,203,424 | 157,800,000 | **-33%** |
| Write | 220,397,869 | 139,994,445 | **-36%** |
| LL cache hits | ~210 | ~50 | **-76%** |

Cold-start (single access) is 3.7-4.2x faster due to no heap allocation on first use.

**Breaking changes** for anyone manually implementing `RegisteredClass` or `PhpClassImpl` (rare, macros handle this). `get_properties()` is gone from the trait. `ClassMetadata::new()` now takes `&'static [PropertyDescriptor<T>]`. The `Property`/`Prop`/`PropertyInfo` types no longer exist.

Net diff: -204 lines.

## Checklist

- [x] I have added a [migration guide](CONTRIBUTING.md#breaking-changes) if applicable.


ptondereau:
#310 for references

PR #722, opened 2026-04-14
fix: use raw-dylib for Windows symbol imports

tob-scott-a:
## Description

Switch from `kind = "dylib"` (the default) to `kind = "raw-dylib"` in the `#[link]` attributes injected into bindgen extern blocks.

With `raw-dylib`, the Rust compiler generates its own import stubs instead of relying on the pre-built import library (`php8.lib`). This ensures that data symbol imports (statics like `zend_ce_exception`, `std_object_handlers`) get proper `__imp_`-prefixed dllimport references. With `kind = "dylib"`, the compiler may emit direct references that neither `rust-lld` nor MSVC `link.exe` can resolve from the import library. Every Zend engine data symbol shows up as an unresolved external.

`raw-dylib` has been stable since Rust 1.65.

## Checklist

### DIFF
--- a/windows_build.rs
+++ b/windows_build.rs
@@ -65,16 +65,25 @@ impl<'a> PHPProvider<'a> for Provider<'a> {
     }
 
     fn write_bindings(&self, bindings: String, writer: &mut impl Write) -> Result<()> {
-        // Symbols don't link without a `#[link(name = "php8")]` attribute on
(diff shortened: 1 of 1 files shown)


tob-scott-a:
This turned out to be unnecessary.

ptondereau:
Unfortunately, this is not the root cause, and I'm currently investigating with a windows server actually

PR #764, opened 2026-08-20
feat: guard get_module export for statically linked extensions

### LINKED ISSUE #763: Guard get_module in static builds to avoid symbol collisions
## Problem

`#[php_module]` always emits a `#[no_mangle] extern "C" fn get_module()`. In a static PHP build this global symbol collides with any other extension that also exports `get_module` (seen in the wild today: a C extension missing the `#ifdef COMPILE_DL_ext` guard linked into the same binary as an ext-php-rs extension).

C extensions avoid this: `ZEND_GET_MODULE` is only compiled for shared builds, so a statically linked C extension exports no `get_module` at all. ext-php-rs has no equivalent guard.

## Proposal

- `#[php_module]` always emits `<crate>_get_module()` (from `CARGO_CRATE_NAME`) holding the real implementation.
- `get_module()` becomes a one-line delegate, emitted unless `EXT_PHP_RS_STATIC_EXT=1` is set at build time. Rust has no `cfg(crate_type)`, so an env var read at macro expansion plays the role of the `COMPILE_DL_ext` guard. Unset keeps today's behavior; dynamic users are unaffected.
- `cargo php static-glue` calls `<crate>_get_module()` instead of `get_module()`. The symbol is derived from the crate name, not `--ext-name`.
- `EXT_PHP_RS_STATIC_EXT=1` also implies `ZEND_ENABLE_STATIC_TSRMLS_CACHE`, replacing `EXT_PHP_RS_STATIC_TSRMLS_CACHE` (introduced in #762, unreleased) so static builds need one switch instead of two.

## Out of scope

Two ext-php-rs extensions in one binary still collide on `ext_php_rs_*` and Rust std symbols. The one-Rust-extension-per-binary limitation stays.

ptondereau:
## Description

Closes #763.

A statically linked ext-php-rs extension always exported `get_module`, so it collided at link time with any other extension exporting the same symbol, like a C extension missing the `#ifdef COMPILE_DL_ext` guard. `#[php_module]` now also exports a crate-prefixed `<crate>_get_module`, and building with `EXT_PHP_RS_STATIC_EXT=1` drops the unprefixed export while keeping the function callable from Rust. The `static-glue` shim calls the prefixed symbol, derived from the library target name rather than `--ext-name`. The variable replaces the unreleased `EXT_PHP_RS_STATIC_TSRMLS_CACHE` and also enables the static TSRMLS cache, so static builds need a single switch. Verified end to end on a two-pass PHP 8.5 ZTS build: linking next to an unguarded C extension failed with `multiple definition of 'get_module'` before, and links and runs with both modules loaded after. CI now asserts the symbol gating with `nm` on the hello_world cdylib.

## Checklist


crates/cli/src/static_glue.rs
```diff
@@ -39,20 +41,21 @@ pub struct StaticGlue {
 
 impl StaticGlue {
     pub fn handle(self) -> AResult<()> {
-        let ext_name = match self.ext_name {
-            Some(name) => name,
-            None => find_staticlib_target(self.manifest.as_deref())?.replace('-', "_"),
-        };
+        let lib_name = find_staticlib_target(self.manifest.as_deref())?.replace('-', "_");
```
okhoshi, line 44: What is the reason to introduce lib name in addition to ext name ?

okhoshi (commented):
Looks good to me, I haven't tried yet however.

But thank you for jumping on so quickly 🚀

crates/cli/src/static_glue.rs
ptondereau: Different things: 

- `lib_name` is what cargo produces and the CLI can't rename: the `lib<target>.a` filename and the `<target>_get_module` symbol the macro derives from `CARGO_CRATE_NAME`.
- `ext_name` is only the php-src identity (`ext/<name>/`, `--enable-<name>`, `<name>_module_entry`) that `--ext-name` may override. 

The old code conflated them, so `--ext-name foo` generated a config.m4 looking for `libfoo.a`, which cargo never creates. That was harmless while the shim called the fixed `get_module`, but with the prefixed symbol, it would be a link error. Without `--ext-name` both are identical.

okhoshi: Got it 👍 more a fix for an oversight of the first PR then.

okhoshi (approved):

PR #765, opened 2026-08-21
feat: introduce --lib-name on static-glue to skip `cargo metadata` call

okhoshi:
## Description

Follow-up on #764, which made the library name always derives from library target name through `cargo metadata`, reintroduce a new flag to skip `cargo metadata` call.

As I'm integrating `ext-php-rs` in a Bazel build, depending on `cargo metadata` is a huge burden: the metadata call refetches the entire dependencies of the crate, because it's ran in isolation. Being able to inject the library name by myself (which I know) makes it lighter. I reckon it's an advanced use case, but I think the entire static-glue already is.

cc @TobiasBengtsson @ptondereau 

## Checklist

### DIFF
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -157,10 +157,15 @@ USAGE:
     cargo-php static-glue [OPTIONS]
 
 OPTIONS:
-        --ext-name <EXT_NAME>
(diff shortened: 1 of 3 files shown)


ptondereau (approved):
LGTM thanks

